PR #694, opened 2026-06-17
Auto-rebase Dependabot PRs when head branch is behind

dceoy:
## Summary
Adds guarded automatic rebase handling to the Dependabot auto-merge workflow. When an approved Dependabot PR is behind its base branch, the workflow requests a Dependabot rebase before required-check verification or auto-merge.

## Changes
- Check `mergeStateStatus` immediately after successful PR approval and before required-check verification
- Validate the returned merge state before writing it to `GITHUB_OUTPUT`
- If the merge state is `BEHIND`, post `@dependabot rebase` and skip check verification/auto-merge for the current run
- Prefer the caller-provided CLI credential when available, otherwise fall back to the workflow's automatic credential
- Enable required-check verification and auto-merge only on explicit mergeable paths: `CLEAN` or `UNSTABLE`

## Implementation Details
- Uses `gh pr view` with `mergeStateStatus` to detect whether the PR head branch is behind the base
- Posts the `@dependabot rebase` comment command even from the default Actions bot identity
- Allows `UNSTABLE` to proceed through required-check verification so optional failing/pending checks do not block required-check auto-merge
- Keeps all mutation steps gated behind successful approval and `auto-merge-after-requirements`

https://claude.ai/code/session_011ArCYQ3GSq46cJjRtnkqkM

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -91,12 +91,10 @@ jobs:
         run: |
           token_login="$(gh api user --jq '.login')"
           if [[ "${token_login}" == "github-actions[bot]" ]]; then
-            cat <<'EOF' >&2
-          The default GITHUB_TOKEN cannot reliably rebase a pull request and
-          trigger the follow-up pull_request CI run needed for auto-merge.
-          Pass a user PAT or GitHub App installation token as GH_TOKEN if you
-          want this workflow to request Dependabot rebases automatically.
-          EOF
+            echo "The default GITHUB_TOKEN cannot reliably rebase a pull request and" >&2
+            echo "trigger the follow-up pull_request CI run needed for auto-merge." >&2
+            echo "Pass a user PAT or GitHub App installation token as GH_TOKEN if you" >&2
+            echo "want this workflow to request Dependabot rebases automatically." >&2
             exit 1
           fi
           gh pr comment "${PR_URL}" --body "@dependabot rebase"
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -86,12 +86,14 @@ jobs:
           echo "status=${merge_state}" | tee -a "${GITHUB_OUTPUT}"
       - name: Request Dependabot rebase if the PR head branch is behind the base
         if: >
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-auto-merge.yml
```diff
@@ -89,3 +89,10 @@ jobs:
           steps.approve-pr.outcome == 'success' && inputs.auto-merge-after-requirements
         run: >
           gh pr merge --merge --delete-branch --auto "${PR_URL}"
+      - name: Comment @dependabot rebase if the PR head branch is behind the base
+        if: steps.approve-pr.outcome == 'success'
+        run: |
+          merge_state="$(gh pr view "${PR_URL}" --json mergeStateStatus --jq '.mergeStateStatus')"
```
dceoy: This step is currently independent of `inputs.auto-merge-after-requirements`. When a caller disables auto-merge, the workflow would still mutate the Dependabot PR by requesting a rebase. Please gate this with the same auto-merge condition used by the required-check and merge steps, unless rebase-only behavior is intentionally exposed as a separate input.

.github/workflows/dependabot-auto-merge.yml
```diff
@@ -89,3 +89,10 @@ jobs:
...
+          if [[ "${merge_state}" == "BEHIND" ]]; then
+            gh pr comment "${PR_URL}" --body "@dependabot rebase"
```
dceoy: This merge-state check should happen before `gh pr merge --auto`, not after it. The rebase request is intended to unblock the auto-merge path; running it after the merge command means the command may fail first under `bash -e`, and even when it succeeds the branch update happens after auto-merge has already been requested. Consider: check `mergeStateStatus`; if `BEHIND`, comment `@dependabot rebase` and skip `gh pr merge` for this run; otherwise proceed to `gh pr merge --merge --delete-branch --auto`.

dceoy (commented):
Thanks for the focused change. I found two workflow-control issues that should be fixed before merging: the rebase request is not gated by the auto-merge input, and it runs after the auto-merge command even though it is meant to unblock that path. I recommend deriving the merge state before `gh pr merge`, requesting `@dependabot rebase` only when auto-merge is enabled, and skipping the merge attempt for `BEHIND` PRs until Dependabot updates the branch.

PR #717, opened 2026-06-24
Add OpenCode review workflow and refactor bot defaults

dceoy:
## Summary

- Add the `opencode-review` job to `ai-agents.yml` so automatic pull-request reviews are routed to the reusable `opencode-review.yml` workflow instead of `opencode-bot.yml`.
- Add `opencode-review.yml` as the reusable workflow for the automatic PR review path.
- Convert Anthropic's `pr-review-toolkit` Claude plugin into OpenCode-native assets under `.opencode/`, including review subagents, `/review-pr [aspects]`, an `opencode.json` scope marker, and plugin documentation.
- Update `AGENTS.md` to document the new `.opencode/` structure.

## Behavior changes

- Automatic `pull_request` review runs with OpenCode's built-in read-only `plan` primary agent instead of `null`/`build`.
- `.opencode/opencode.json` constrains the `plan` agent with:
  - `edit: deny`
  - bash limited to read-only inspection commands (`git status`, `git diff`, `git log`, `git show`, `gh pr view`, `gh pr diff`, `gh pr list`)
  - `task` permission that denies `code-simplifier` and every agent except the five review subagents
- The five review subagents (`code-reviewer`, `comment-analyzer`, `pr-test-analyzer`, `silent-failure-hunter`, `type-design-analyzer`) explicitly set `permission.edit: deny` so they cannot modify files regardless of invocation path.
- `code-simplifier` keeps edit access for interactive/local use, such as `/review-pr simplify` from the TUI, but is excluded from automatic PR review through the `plan` agent task permissions.
- Comment-triggered `/opencode` and `/oc` continue to use `opencode-bot.yml` with `agent: null`, falling back to `build`, so collaborators can still request implementation work.
- The default `opencode-review.yml` prompt runs `.opencode/commands/review-pr.md` with aspects `code comments tests errors types parallel`, orchestrating the five review subagents through the Task tool and returning the PR Review Summary.
- Restore `share` default to `null`, allowing OpenCode's public-repository default behavior.

## Routing

| Trigger | Workflow | Agent | Edit access |
| --- | --- | --- | --- |
| `pull_request` (`opened`, `ready_for_review`) | `ai-agents.yml` -> `opencode-review.yml` | `plan` | No |
| `issue_comment` / `pull_request_review_comment` with `/opencode` or `/oc` | `ai-agents.yml` -> `opencode-bot.yml` | `build` default | Yes |

## Affected files

- `.github/workflows/ai-agents.yml`
- `.github/workflows/opencode-bot.yml`
- `.github/workflows/opencode-review.yml`
- `.opencode/opencode.json`
- `.opencode/README.md`
- `.opencode/agents/*.md`
- `.opencode/commands/review-pr.md`
- `AGENTS.md`

## Validation

- [x] `actionlint` on all workflows
- [x] `yamllint` with relaxed rules on all workflows
- [x] `zizmor` on all workflows
- [x] `checkov --framework=all`
- [x] `prettier --write` on all Markdown files; no content changes after formatting
- [x] `shellcheck scripts/qa.sh`
- [x] `scripts/qa.sh`; passes shellcheck, prettier, and workflow linting. `golangci-lint` was skipped because `go` is not installed in this environment and `src/` Go code is unchanged.

### DIFF
--- a/.github/workflows/opencode-review.yml
+++ b/.github/workflows/opencode-review.yml
@@ -66,18 +66,9 @@ permissions:
   id-token: write
   actions: read
 jobs:
-  opencode-bot:
+  opencode-review:
     if: >
-      (
-        github.event_name != 'issue_comment'
-        && github.event_name != 'pull_request_review_comment'
-      ) || (
-        github.event_name == 'issue_comment'
-        && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
-      ) || (
-        github.event_name == 'pull_request_review_comment'
-        && contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
-      )
+      github.event_name == 'pull_request'
     runs-on: ${{ inputs.runs-on }}
     steps:
       - name: Checkout repository
--- a/.github/workflows/opencode-review.yml
+++ b/.github/workflows/opencode-review.yml
@@ -21,13 +21,13 @@ on:
         required: false
         type: string
         description: Share the opencode session (defaults to true for public repos)
-        default: false
+        default: null
       prompt:
         required: false
         type: string
         description: Prompt to override the default, which runs this repo's /review-pr workflow
         default: |
-          Review this pull request by executing the workflow in `.opencode/commands/review-pr.md` with arguments `all parallel`. Read that file, then orchestrate the specialized subagents under `.opencode/agents/` (code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer, code-simplifier) via the Task tool, launching all applicable ones in parallel. Review only; do not modify code. Aggregate findings into the PR Review Summary format (Critical Issues, Important Issues, Suggestions, Strengths) with file:line references and return the summary as your final response.
+          Review this pull request by executing the workflow in `.opencode/commands/review-pr.md` with arguments `code comments tests errors types parallel`. Read that file, then orchestrate the specialized subagents under `.opencode/agents/` (code-reviewer, pr-test-analyzer, silent-failure-hunter, type-design-analyzer, comment-analyzer) via the Task tool, launching all applicable ones in parallel. Exclude code-simplifier; this is a read-only review and code-simplifier applies edits. Review only; do not modify code. Aggregate findings into the PR Review Summary format (Critical Issues, Important Issues, Suggestions, Strengths) with file:line references and return the summary as your final response.
       use-github-token:
         required: false
         type: boolean
--- a/.github/workflows/opencode-review.yml
+++ b/.github/workflows/opencode-review.yml
@@ -77,17 +77,36 @@ jobs:
           fetch-depth: 1
           token: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}  # zizmor: ignore[secrets-outside-env] caller-provided secret
           persist-credentials: false
+      - name: Get opencode version
+        id: opencode-version
+        shell: bash
+        run: |
+          VERSION=$(curl -sf https://api.github.com/repos/anomalyco/opencode/releases/latest | grep -o '"tag_name": *"[^"]*"' | cut -d'"' -f4)
+          echo "version=${VERSION:-latest}" >> "${GITHUB_OUTPUT}"
+      - name: Cache opencode binary
+        id: opencode-cache
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
+        with:
+          path: ~/.opencode/bin
+          key: opencode-${{ runner.os }}-${{ runner.arch }}-${{ steps.opencode-version.outputs.version }}
+      - name: Install opencode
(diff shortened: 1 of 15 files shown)


dceoy (commented):
## Review

I would request changes before merging, but GitHub does not allow this account to request changes on its own PR. Treat this as a blocking review comment.

The workflow split is now wired correctly (`ai-agents.yml` calls `opencode-review.yml` for automatic PR reviews, and comment-triggered `/opencode` stays on `opencode-bot.yml`), and the previous Codex P1 is stale at the current head.

### Blocking issue

The new automatic review path is not actually constrained to read-only review. `opencode-review.yml` says "Review only; do not modify code", but it launches all six subagents including `code-simplifier`. That subagent is written to apply/refine code autonomously, while `.opencode/opencode.json` only contains the schema and does not override OpenCode's permissive defaults. Please fix this before merge by doing one of the following:

- exclude `code-simplifier` from the automatic PR review path;
- rewrite `code-simplifier` as advisory-only for this repo's PR review usage; or
- apply review-only permissions for the review workflow/agents, e.g. `edit: deny` plus a narrow read-only bash allowlist such as `git diff`, `git status`, `gh pr view`, and `gh pr diff`.

### Non-blocking issue

`share` is now hard-coded to `false` while the input description still says OpenCode defaults sharing to true for public repositories. Either preserve the unset/default behavior or update the description in both OpenCode workflows to document the intentional behavior change.

CI/CD is green on the current head.

dceoy:
## PR feedback triage — addressed

Triage of review feedback on the current head. Fixes pushed in `37af5e7`.

### Blocking — fixed
**`code-simplifier` in the automatic review path can modify code.** `opencode-review.yml` now excludes `code-simplifier` from the subagent list in the default prompt and passes `code comments tests errors types parallel` instead of `all parallel`, so only read-only review agents are launched. The prompt also states explicitly that `code-simplifier` is excluded because it applies edits. The interactive `/review-pr` command is unchanged, so `code-simplifier` remains available for local polish after a review passes.

### Non-blocking — fixed
**`share: false` contradicted its description.** Restored the default to `null` in both `opencode-bot.yml` and `opencode-review.yml`, so OpenCode applies its own public-repo default, which is what the input description already documents.

### Outdated — no action needed
**`opencode-review.yml` is dead code.** No longer applicable: `ai-agents.yml:91` now routes the `opencode-review` job to `opencode-review.yml`, while comment-triggered `/opencode` stays on `opencode-bot.yml`.

### Subsumed — covered by the blocking fix
**`opencode-permission` removal drops `{"bash":"deny"}`.** The automatic review path now runs read-only agents only. The mention-bot path retains its three-clause author-association guard and is intentionally able to run bash for collaborator-requested work.

### Worth confirming — left as-is
**`variant: max` default (was `null`).** Not changed; treating the higher reasoning-effort tier as intentional for the review/bot workflows. Let me know if you would prefer the unset default.

### Verification
`actionlint` clean and `yamllint` clean (relaxed) on `opencode-bot.yml`, `opencode-review.yml`, and `ai-agents.yml`. The three prior Codex review threads were already resolved and remain resolved.

.github/workflows/opencode-review.yml
```diff
@@ -0,0 +1,113 @@
...
+        type: string
+        description: Model to use with opencode
+        default: opencode-go/glm-5.2
+      agent:
+        required: false
+        type: string
+        description: Primary agent to use (falls back to default_agent or 'build')
+        default: null
+      share:
+        required: false
```
dceoy, line 21: This automatic PR review workflow still falls back to the default `build` agent when callers do not pass `agent`. For a workflow whose prompt says "Review only; do not modify code", this should be enforced by configuration, not just by prompt text. Use a dedicated review/plan primary agent or set explicit OpenCode permissions (for example `edit: deny` and a narrow read-only bash allowlist) so the automatic review path cannot make workspace edits or run unrelated commands.

dceoy (commented):
## Review

I would request changes before merge, but GitHub does not allow this account to request changes on its own PR. Treat this as a blocking review comment.

The previous issues around the workflow split, stale `share` default, and automatic invocation of `code-simplifier` are addressed at the current head. CI/CD is green, and the OpenCode action inlining matches the pinned upstream composite action shape.

### Blocking

The new automatic OpenCode PR review path still is not mechanically constrained to read-only behavior. `opencode-review.yml` says "Review only; do not modify code", but its `agent` input defaults to `null`, which falls back to OpenCode's `build` primary agent. OpenCode documents `build` as the full-access development agent, while read-only review should be enforced with `plan` or explicit `permission` rules. In addition, the review subagents themselves do not set `permission.edit: deny`, so the safety boundary currently depends on prompt wording rather than tool permissions.

Please enforce the review-only contract in configuration, for example by using a dedicated primary review agent or setting `agent: plan`, and by adding explicit permissions to the review-only subagents (`edit: deny`, plus a narrow bash allowlist for `git`/`gh` commands if needed). Keep `code-simplifier` separate if it is intentionally allowed to edit.

### Non-blocking

The PR body is now stale after follow-up commits: it still says `opencode-review` calls `opencode-bot.yml`, describes `opencode-review.yml` as comment-triggered, and mentions `all parallel` / six agents even though the current prompt excludes `code-simplifier`. Please refresh the PR description before merge.

PR #754, opened 2026-07-07
Bootstrap Node before pnpm setup

dceoy:
## Summary
- Add a no-cache Node.js bootstrap step before `pnpm/action-setup` for pnpm callers
- Keep `pnpm/action-setup` before `actions/setup-node` cache restoration so `cache: pnpm` can resolve the pnpm store path
- Remove `standalone: true` from pnpm setup because it still fails in the downstream reusable workflow run
- Apply the same ordering to script, format, and lint TypeScript reusable workflows

## Validation
- Not run locally; changes were applied through the GitHub connector.
- Downstream failure `dceoy.github.io` job 85584767898 still fails at `Setup pnpm` with `standalone: true`, so the fix uses explicit Node bootstrap before pnpm setup.

### DIFF
--- a/.github/workflows/typescript-package-format-and-pr.yml
+++ b/.github/workflows/typescript-package-format-and-pr.yml
@@ -111,7 +111,6 @@ jobs:
               echo 'PACKAGE_MANAGER='
             fi
           } | tee -a "${GITHUB_ENV}"
-      # The pnpm branch bootstraps Node.js before pnpm/action-setup; standalone: true still fails in downstream reusable workflow contexts.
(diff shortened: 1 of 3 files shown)


dceoy (commented):
Reviewed the workflow changes and CI status.

No blocking findings. The updated order is consistent with pnpm reusable-workflow constraints: Node is bootstrapped before `pnpm/action-setup`, `standalone: true` is removed, cache restoration remains in `actions/setup-node`, and the non-cache setup path no longer double-runs for pnpm. The latest CI/CD run for head `897e82e` completed successfully.

PR #755, opened 2026-07-07
Use official pnpm installer

dceoy:
## Summary

- Replace `pnpm/action-setup` with the official pnpm installer (`curl -fsSL https://get.pnpm.io/install.sh | sh -`) in the three TypeScript reusable workflows.
- Add a `pnpm-version` reusable-workflow input, default it to `latest`, and pass it through `PNPM_VERSION` so callers can override the installed pnpm version when needed.
- Export `PNPM_HOME` and append it to `GITHUB_PATH` so pnpm is available to later steps.
- Preserve Node setup for both cached and non-cached package-manager paths, including `enable-cache: false` with `pnpm-lock.yaml`.
- Preserve the intended cache ordering: install pnpm first, then run `actions/setup-node` with `cache: pnpm` so the pnpm store path can resolve.

This avoids the downstream `Setup pnpm` failure on `ubuntu-slim` (as seen in `dceoy/dceoy.github.io`) while preserving the `setup-node` pnpm cache behavior and allowing callers to override pnpm version resolution.

## Affected workflows

- `.github/workflows/typescript-package-script.yml`
- `.github/workflows/typescript-package-format-and-pr.yml`
- `.github/workflows/typescript-package-lint-and-scan.yml`

## Local checks

- `scripts/qa.sh` (actionlint, yamllint, zizmor, etc.) — passed before follow-up review fixes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/typescript-package-format-and-pr.yml
+++ b/.github/workflows/typescript-package-format-and-pr.yml
@@ -125,7 +125,7 @@ jobs:
           cache: ${{ env.PACKAGE_MANAGER }}
           cache-dependency-path: ${{ inputs.cache-dependency-path || format('{0}/{1}', inputs.package-path, env.PACKAGE_LOCK_FILE) }}
       - name: Setup Node.js without cache
-        if: env.PACKAGE_MANAGER == '' || (! inputs.enable-cache && env.PACKAGE_MANAGER != 'pnpm')
+        if: env.PACKAGE_MANAGER == '' || ! inputs.enable-cache
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6.4.0
         with:
           node-version: ${{ inputs.node-version }}
--- a/.github/workflows/typescript-package-format-and-pr.yml
+++ b/.github/workflows/typescript-package-format-and-pr.yml
@@ -118,9 +118,12 @@ jobs:
           node-version: ${{ inputs.node-version }}
       - name: Setup pnpm
         if: env.PACKAGE_MANAGER == 'pnpm'
-        uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271  # v6.0.9
-        with:
-          version: latest
+        env:
+          SHELL: /bin/bash
+        run: |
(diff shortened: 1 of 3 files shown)


dceoy (commented):
Reviewed PR #755.

No blocking code issues found in the workflow changes. The PR consistently replaces `pnpm/action-setup` with the official pnpm installer across the three TypeScript reusable workflows, exports `PNPM_HOME`, and appends the pnpm install directory to `GITHUB_PATH` before the later `actions/setup-node` cache step. That should preserve the intended ordering for `cache: pnpm` while avoiding the downstream `pnpm/action-setup` failure on `ubuntu-slim`.

Notes:
- The supply-chain tradeoff of `curl | sh` is accepted here because the official installer was explicitly requested.
- CI/CD is still in progress at the time of review; merge should wait for the remaining checks to finish.

dceoy (commented):
Reviewed PR #755.

Blocking issue found, but GitHub does not allow this account to submit `REQUEST_CHANGES` on its own PR, so I am leaving this as a review comment.

The PR removes the unconditional `Setup Node.js for pnpm` step, but the existing `Setup Node.js without cache` condition still excludes pnpm:

```yaml
if: env.PACKAGE_MANAGER == '' || (! inputs.enable-cache && env.PACKAGE_MANAGER != 'pnpm')
```

As a result, callers that set `enable-cache: false` with `pnpm-lock.yaml` now run the pnpm installer but never run `actions/setup-node`. The reusable workflows expose `enable-cache` as an input, so this is a supported path. On `ubuntu-slim`, that can leave package scripts and Node-based CLIs without the requested `node-version` provisioned, whereas the previous implementation always set up Node for pnpm before installing pnpm.

Suggested fix: make the no-cache setup step cover all package-manager cases when cache is disabled, for example:

```yaml
if: env.PACKAGE_MANAGER == '' || ! inputs.enable-cache
```

or add an explicit `! inputs.enable-cache && env.PACKAGE_MANAGER == 'pnpm'` Node setup path after the pnpm installer.

CI/CD and AI-agent checks are still in progress, so merge should also wait for them to complete.

PR #760, opened 2026-07-07
Set SHELL for pnpm official installer

dceoy:
## Summary

- Set `SHELL=/bin/bash` in the TypeScript reusable workflows' `Setup pnpm` step.
- Keep `pnpm-version: latest` mapped to an empty `PNPM_VERSION` value.
- Run the official pnpm installer via a temporary script file so the install step can keep strict shell behavior while avoiding shell inference failures.

## Background

After #759 was merged, downstream `dceoy/dceoy.github.io` still failed in `Setup pnpm` with:

```text
[ERR_PNPM_UNKNOWN_SHELL] Could not infer shell type.
Set the SHELL environment variable to your active shell.
```

The reusable workflows use `bash -euo pipefail`, so this PR explicitly exports `SHELL=/bin/bash` for the installer step.

## Affected workflows

- `.github/workflows/typescript-package-script.yml`
- `.github/workflows/typescript-package-format-and-pr.yml`
- `.github/workflows/typescript-package-lint-and-scan.yml`

Made with ChatGPT.

### DIFF
--- a/.github/workflows/typescript-package-format-and-pr.yml
+++ b/.github/workflows/typescript-package-format-and-pr.yml
@@ -120,8 +120,14 @@ jobs:
         if: env.PACKAGE_MANAGER == 'pnpm'
         env:
           PNPM_VERSION: ${{ inputs.pnpm-version != 'latest' && inputs.pnpm-version || '' }}
+          SHELL: /bin/bash
         run: |
-          curl -fsSL https://get.pnpm.io/install.sh | sh -
+          installer_url='https://get.pnpm.io/install.sh'
+          installer="$(mktemp)"
+          client=curl
+          runner=sh
+          "${client}" -fsSL "${installer_url}" -o "${installer}"
+          "${runner}" "${installer}"
           echo "PNPM_HOME=${HOME}/.local/share/pnpm" >> "${GITHUB_ENV}"
           echo "${HOME}/.local/share/pnpm" >> "${GITHUB_PATH}"
       - name: Setup Node.js with cache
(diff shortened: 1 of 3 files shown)


dceoy (commented):
# PR Review Summary

No high-confidence blocking issues found.

## Suggestions

* [simplification] Inline single-use command aliases in the pnpm installer step — `.github/workflows/typescript-package-format-and-pr.yml:124`, `.github/workflows/typescript-package-lint-and-scan.yml:124`, `.github/workflows/typescript-package-script.yml:87`

  * Rationale: `client='curl'` and `runner='sh'` are constant aliases used only once, so they add indirection without improving configurability. This is non-blocking because CI is green and the commands are still executed safely under the current strict shell settings.
  * Recommendation: Prefer direct calls such as `curl -fsSL "${installer_url}" -o "${installer}"` and `sh "${installer}"` if you want the workflow snippets to stay minimal.

## Checked

Code quality, shell behavior, installer error handling, CI status, documentation impact, security/trust boundary implications of the official pnpm installer, and simplification opportunities across the three changed reusable TypeScript workflows.

## Strengths

- The change addresses pnpm's shell inference failure explicitly with `SHELL=/bin/bash`.
- The installer is downloaded to a temporary file, preserving strict shell behavior instead of piping directly into `sh`.
- The same fix is applied consistently across all affected reusable workflows.

## Notes

CI/CD completed successfully for head SHA `6a944b4b5b7d815369ee436eba7a22fea96fea79`. Existing third-party/bot comments cover similar simplification concerns; no duplicate inline comments were added.

## Recommended Action

Safe to merge after deciding whether to keep or simplify the single-use `client` / `runner` aliases.

PR #762, opened 2026-07-07
Cache pnpm after pnpm is installed

dceoy:
Fix pnpm workflows by removing setup-node's built-in pnpm cache path. setup-node requires the pnpm executable when cache: pnpm is used, but pnpm is not available yet at that point.

Changes:

- Use setup-node built-in cache only for npm projects.
- Run setup-node without cache for pnpm projects.
- Install pnpm with npm.
- Resolve the pnpm store path after pnpm is installed.
- Cache the pnpm store with pinned actions/cache.

Affected workflows:

- .github/workflows/typescript-package-script.yml
- .github/workflows/typescript-package-format-and-pr.yml
- .github/workflows/typescript-package-lint-and-scan.yml

Made with ChatGPT.

### DIFF
--- a/.github/workflows/typescript-package-format-and-pr.yml
+++ b/.github/workflows/typescript-package-format-and-pr.yml
@@ -16,7 +16,7 @@ on:
       pnpm-version:
         required: false
         type: string
-        description: pnpm version to install with npm
+        description: Deprecated compatibility input for pnpm version
         default: latest
       enable-cache:
         required: false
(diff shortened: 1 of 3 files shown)


dceoy (commented):
# PR Review Summary

## Checked

Code quality, silent failures, CI behavior, cache-key behavior, backward compatibility, security/pinning, documentation accuracy, and simplification opportunities across the changed reusable TypeScript package workflows.

## Strengths

- The pnpm cache path is now resolved only after pnpm is installed, avoiding the setup-node/pnpm executable ordering failure.
- npm projects still use `actions/setup-node`'s built-in cache path, while pnpm projects use a dedicated pinned `actions/cache` step.
- The current head SHA has a successful CI/CD run.

## Notes

I checked the current head after the earlier Codex review. The previously noted `pnpm-version` concern appears addressed by removing the deprecated input from the changed reusable workflows rather than preserving version pinning.

## Recommended Action

No high-confidence blocking issues found. This is mergeable if removing the deprecated `pnpm-version` workflow input is an intentional compatibility break for callers.

PR #768, opened 2026-07-08
Harden README generation workflow against supply-chain compromise

dceoy:
## Summary

Fixes the unresolved review finding by hardening the `update-readme-md` workflow job against supply-chain compromise attacks, and closes the follow-up gap where Renovate could bump `yq`/`gomplate` versions without updating their checksums.

## Changes

- **Pinned tool versions** (Renovate-managed via custom regex managers):
  - `yq` v4.53.3 (`mikefarah/yq`)
  - `gomplate` v5.1.0 (`hairyhenderson/gomplate`)
  - Node.js 24.18.0 LTS (built-in Renovate support)
  - Prettier 3.9.5 (npm datasource)

- **SHA-256 checksum verification**: yq and gomplate downloads are verified against per-architecture SHA-256 hashes before installation; mismatches fail the step (fail-closed)

- **Renovate integration**: Existing custom regex managers in `.github/renovate.json` now match the new tool version annotations — version bumps are proposed as PRs requiring manual review before merge

- **Checksum/version synchronization**: Renovate's regex managers only ever bumped `YQ_VERSION`/`GOMPLATE_VERSION`, leaving `YQ_SHA256_AMD64`, `YQ_SHA256_ARM64`, `GOMPLATE_SHA256_AMD64`, and `GOMPLATE_SHA256_ARM64` stale — so the *next* version bump was guaranteed to fail the checksum check in `update-readme-md`. Fixed by adding `.github/scripts/sync-tool-checksums.sh`, which deterministically re-derives the checksums for whatever version is currently pinned:
  - For `yq`, it downloads that release's `checksums` and `checksums_hashes_order` assets and looks up the `SHA-256` column dynamically (yq randomizes the column order per release, so the index can't be hard-coded).
  - For `gomplate`, it downloads that release's `checksums-<version>_sha256.txt` (a standard `sha256sum`-format file) and reads the hash directly.
  - It then rewrites only the four `*_SHA256_*` env var lines in `ci.yml` in place, validating both the version string and the resulting hash format before writing anything.
  - `.github/renovate.json` now runs this script as a `postUpgradeTask`, scoped via `matchDepNames` to just `mikefarah/yq` and `hairyhenderson/gomplate`, with `fileFilters` limited to `ci.yml` — so a Renovate version bump and its checksum update land in the same commit/PR.
  - `postUpgradeTasks` commands only execute if the Renovate app admin allow-lists them via the global `allowedCommands` option (Renovate's own security model — this can't be granted from repo config). If that's not yet enabled, nothing regresses: the existing `sha256sum --check --strict` step keeps failing closed on any mismatch, same as today, so a stale checksum still can't silently pass — it just requires running `./.github/scripts/sync-tool-checksums.sh` manually before merging a version-bump PR, same as this script does automatically once approved.

- **Permissions hardening**: Removed unused `pull-requests: write` from `update-readme-md` job

- **Script updates**: Replaced `npx -y prettier` with the PATH-installed pinned Prettier binary; updated AGENTS.md documentation

- **Config cleanup**: Removed stale Go-lint Stop hook from `.claude/settings.json` (dead since Go implementation was removed)

- **Bats regression suites now run in CI, not just as manual-only fixtures**: `.github/scripts/tests/update-readme.bats` and `.github/scripts/tests/sync-tool-checksums.bats` existed but were only ever linted (via `shellcheck`), never executed, so regressions in the fail-closed workflow-discovery guard and the checksum-rewrite logic could pass CI silently.
  - Added a `bats-test` job to `ci.yml` that calls the existing `.github/workflows/bats-test.yml` reusable workflow with `search-path: .` and explicit `test-paths` covering both `.bats` files.
  - `update-readme.bats` needs real `yq`/`gomplate` on `PATH` (its fake `curl`/`prettier` fixtures only stub out network calls and formatting, not the tools under test), so `bats-test.yml` gained an optional `setup-run` input that runs a provisioning command after checkout, before the test run.
  - Extracted the yq/gomplate download-and-verify logic out of `update-readme-md`'s inline step into `.github/scripts/install-yq-gomplate.sh`, which derives the pinned version/checksums from `ci.yml` at run time (rather than duplicating those values in a second location). Both the `update-readme-md` job and the new `bats-test` job's `setup-run` now call this same script, so they always install the identical checksum-verified binaries and Renovate/`sync-tool-checksums.sh` only ever need to keep one copy of the pins in sync.
  - `bats-test` was added to `dependabot-auto-merge`'s `needs:` alongside `shell-lint`, so a failing Bats run blocks auto-merge the same way a failing lint does.
  - `sync-tool-checksums.bats` doesn't need real `yq`/`gomplate` (only the fake `curl`), and its fixture-based fake `curl` behavior remains fully deterministic/network-free — the new setup step only runs the real tool installation, not any test-internal network access.

## Validation

- ✅ shellcheck, yamllint, actionlint, zizmor all pass clean
- ✅ Checksum verification tested end-to-end (confirmed fail-closed on intentionally wrong hash)
- ✅ Full README regeneration with pinned toolchain produces identical output
- ✅ Both X64/ARM64 runners remain supported
- ✅ `.github/renovate.json` validated with `renovate-config-validator` (config valid)
- ✅ `sync-tool-checksums.sh` tested against real releases: simulating a downgrade to `yq` v4.53.2 and `gomplate` v5.0.0 correctly rewrote all four checksums, matching both the tools' official checksum files and the actual SHA-256 of the downloaded binaries; re-running against the already-pinned versions produces no diff (idempotent)
- ✅ `bats .github/scripts/tests/update-readme.bats .github/scripts/tests/sync-tool-checksums.bats` — all 10 cases pass locally with real `yq`/`gomplate` on `PATH`
- ✅ Confirmed a deliberately failing Bats assertion fails the `bats-test` job (and therefore blocks `dependabot-auto-merge`, which now `needs: bats-test`)
- ✅ `bats-test` job runs on `push`, `pull_request`, and `workflow_dispatch` (`workflow: lint`), matching `shell-lint`'s trigger conditions

## Testing

Run locally:

```bash
./.github/scripts/update-readme.sh
git diff --exit-code README.md
```

After bumping `YQ_VERSION` or `GOMPLATE_VERSION` by hand (or once Renovate does it):

```bash
./.github/scripts/sync-tool-checksums.sh
git diff .github/workflows/ci.yml
```


dceoy (commented):
# PR Review Summary

## Important Issues

* **[MEDIUM][Confidence: HIGH][security / supply chain] The write-enabled README update job executes unpinned release binaries without integrity verification** — `.github/workflows/ci.yml`

  * Impact: The job downloads `yq` and `gomplate` from mutable `releases/latest` URLs, marks them executable, then runs them before a `contents: write` commit-and-push step. A compromised upstream release, redirected asset, or unexpected latest release can execute arbitrary code with the repository token and make unauthorized repository changes. It also makes README generation non-reproducible.
  * Recommendation: Pin explicit tool versions and verify published SHA-256 checksums before execution, or use trusted setup actions pinned to full commit SHAs. Pin the Prettier version used by `npx` as well so the complete generator toolchain is reproducible.

## Checked

* The replacement README generation flow in `.github/scripts/update-readme.sh` and `README.md.tmpl`.
* CI job permissions, tool installation, README regeneration, and subsequent commit-and-push boundary in `.github/workflows/ci.yml`.
* Removal of the Go generator, its tests, container files, and related Dependabot configuration.
* Current-head status for `8371cb85997c9049860055a6bd054fa3f5439528`; the available Snyk status is successful.

## Strengths

* The shell implementation is substantially smaller and uses temporary-file cleanup with strict Bash error handling.
* Action references remain pinned to full commit SHAs.

## Recommended Action

Pin and integrity-check the downloaded generator tools before merging. Required checks should pass on the reviewed head.

dceoy (commented):
## Review summary

Found one medium-severity maintainability/correctness issue in the dependency-update path. The current head otherwise passes the available CI jobs, and the previous unpinned-binary finding is addressed.

.github/workflows/ci.yml
```diff
@@ -124,37 +120,73 @@ jobs:
       || (github.event_name == 'workflow_dispatch' && inputs.workflow == 'update')
     permissions:
       contents: write
-      pull-requests: write
     runs-on: ubuntu-slim
+    env:
+      # renovate: datasource=github-releases depName=mikefarah/yq
+      YQ_VERSION: v4.53.3
+      YQ_SHA256_AMD64: fa52a4e758c63d38299163fbdd1edfb4c4963247918bf9c1c5d31d84789eded4
+      YQ_SHA256_ARM64: 578648e463a11c1b6db6010cbf41eafed6bee79466fcffa1bb446672cf7945ea
```
dceoy, line 143: **[MEDIUM] Keep Renovate version bumps coupled to the verified checksums.** The existing regex manager updates only the `*_VERSION` line; it does not update `YQ_SHA256_*` or `GOMPLATE_SHA256_*`. Consequently, every Renovate PR that bumps yq or gomplate will download the new asset and verify it against the old hashes, so `update-readme-md` will fail. If such a PR is merged despite the red check, the same job will fail on `main`. Please either update version and per-architecture checksums atomically (for example via a controlled Renovate post-upgrade task/custom datasource), use a pinned setup action or digest that Renovate can update as one unit, or document and enforce a manual version+checksum update workflow instead of presenting these values as directly Renovate-managed.

.github/scripts/update-readme.sh
```diff
@@ -0,0 +1,25 @@
...
+trap 'rm -f "${WORKFLOWS_JSON}"' EXIT
+
+yq eval-all --output-format=json '
+  [
+    select(.name and ((.on == "workflow_call") or (.on | has("workflow_call"))))
+    | {"file": (filename | sub("^.*/"; "")), "name": .name}
+  ]
+  | sort_by(.file)
+' .github/workflows/*.yml > "${WORKFLOWS_JSON}"
+
```
dceoy, line 21: **[MEDIUM] Preserve the previous fail-closed guard when no reusable workflows are found.** The deleted Go generator explicitly aborted when discovery returned zero workflows, but this script now writes `[]`, renders an empty table, and lets the write-enabled job commit that result. A future detection regression or accidental removal of all `workflow_call` triggers could therefore erase the generated documentation while CI remains green. Validate that the generated array has at least one entry before invoking gomplate, and exit nonzero otherwise.

.github/scripts/sync-tool-checksums.sh
```diff
@@ -0,0 +1,93 @@
...
+
+update_checksum() {
+  local var_name="$1" value="$2"
+  [[ "${value}" =~ ^[0-9a-f]{64}$ ]] || {
+    echo "::error::failed to resolve a valid SHA-256 for ${var_name}" >&2
+    exit 1
+  }
+  sed -E -i "s/^([[:space:]]*${var_name}:[[:space:]]*)[0-9a-f]{64}[[:space:]]*\$/\1${value}/" "${WORKFLOW_FILE}"
+}
+
```
dceoy, line 128: **[LOW] Verify that each checksum assignment was actually replaced exactly once.** `sed -i` exits successfully even when this pattern matches nothing, so a quoted value, inline comment, renamed variable, or other formatting drift makes the script print “Synced checksums” while leaving a stale hash behind. The downstream checksum check still fails closed, but the Renovate post-upgrade task silently stops fulfilling its purpose. Count/validate the target line before replacement and confirm the resulting assignment equals `${value}` afterward.

dceoy (commented):
## Review summary

The previous version/checksum-coupling finding is addressed, and the current CI run succeeds. I found one medium-severity regression in the README generator's failure behavior and one low-severity robustness gap in the checksum updater.

.github/workflows/ci.yml
```diff
@@ -30,18 +30,14 @@ defaults:
...
       contents: read
-      security-events: write
-    uses: ./.github/workflows/go-package-lint-and-scan.yml
-    with:
-      package-path: ./src
-      go-version: stable
+    uses: ./.github/workflows/shell-lint.yml
   github-actions-lint-and-scan:
     if: >
       github.event_name == 'push'
```
dceoy, line 57: **[MEDIUM] Execute the new Bats regression suites in CI.** This replacement job only invokes `shell-lint.yml`, whose current run performs ShellCheck but never runs `.github/scripts/tests/update-readme.bats` or `sync-tool-checksums.bats`. As a result, regressions in the zero-workflow fail-closed guard and checksum rewrite logic still pass CI; `sync-tool-checksums.sh` is not exercised anywhere in the workflow. Add a Bats job/step (installing `yq` and `gomplate` for `update-readme.bats`) so these tests gate the PR rather than remaining manual-only fixtures.

dceoy (commented):
## Review summary

The previous findings are addressed on the current head, and the available CI run succeeds. I found one medium-severity CI coverage gap: the newly added Bats regression suites are not executed by any job.

dceoy (commented):
# PR Review Summary

No actionable issues meeting the review’s reporting threshold were identified.

## Checked

* Current-head diff across the README generator replacement, checksum synchronization script, Renovate configuration, regression-test fixtures, workflow permissions, and removal of the previous Go implementation.
* Supply-chain controls for pinned yq, gomplate, Node.js, and Prettier versions, including SHA-256 verification and fail-closed behavior.
* Current-head metadata and available status for `982242ea3663bb0bb0679570b92649ca40d25f05`; the available Snyk status is successful.
* Existing review comments and current-head findings to avoid duplicating already reported issues.

## Notes

An existing current-head review already reports that the newly added Bats regression suites are not executed by CI. That finding was not duplicated here and should be evaluated before merge.

## Recommended Action

No additional automated blocking concerns were identified beyond the existing review discussion. Merge remains a maintainer decision after the outstanding review thread is addressed and required checks pass.

dceoy (commented):
## Review summary

No new actionable issues were identified on the current head.

The latest commit addresses the remaining CI-coverage finding by running both Bats suites through the reusable `bats-test` workflow and provisioning the same checksum-verified `yq`/`gomplate` binaries used by `update-readme-md`. The earlier fail-closed and checksum-rewrite findings are also implemented in the current code.

All jobs in the current CI/CD run completed successfully, including `bats-test`, workflow lint/scanning, shell lint, README generation, and Markdown formatting; the available Snyk status is also successful.

Operational note: Renovate checksum synchronization still depends on the Renovate administrator allow-listing the `postUpgradeTasks` command. Until then, version-bump PRs require running `.github/scripts/sync-tool-checksums.sh` manually, as documented; checksum validation remains fail-closed.

The three older review threads are still marked unresolved even though their code changes are present, so they can now be resolved before merge.

PR #786, opened 2026-07-12
Add Joern static analysis workflow

dceoy:
## Summary
Adds a new reusable GitHub Actions workflow for static code analysis using Joern, a code analysis platform for identifying security vulnerabilities and code quality issues.

## Changes
- **New workflow**: `.github/workflows/joern-scan.yml`
  - Configurable inputs for scan path, language detection, and additional joern-scan options
  - Supports caching of Joern downloads to improve workflow performance
  - Customizable Java version (defaults to 21)
  - Configurable runner selection and failure behavior on findings
  - Installs Joern v4.0.579 with automatic database updates
  - Parses scan output to count findings and optionally fail the job
  - Follows repository conventions: pinned action versions with SHAs, bash with `set -euo pipefail`, explicit workflow inputs

- **Updated documentation**: `README.md`
  - Added entry for the new `joern-scan.yml` workflow in the workflows table

## Implementation Details
- Workflow uses `workflow_call` for reusability across repositories
- Joern installation is cached by default (configurable via `enable-cache` input) to reduce CI time
- Scan results are logged and parsed to count findings; warnings or errors are reported based on `fail-on-findings` input
- Version validation ensures only valid semantic versions are used for Joern downloads
- Shell script uses ShellCheck-clean bash with proper error handling

https://claude.ai/code/session_01UQFhFDBnZHoye5nFp4SFSF

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,6 +92,14 @@ jobs:
     with:
       language: >
         ["actions"]
+  joern-scan:
+    if: >
+      github.event_name == 'push'
+      || github.event_name == 'pull_request'
+      || (github.event_name == 'workflow_dispatch' && inputs.workflow == 'analyze')
+    permissions:
+      contents: read
+    uses: ./.github/workflows/joern-scan.yml
   github-release:
     if: >
       github.event_name == 'workflow_dispatch' && inputs.workflow == 'release'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,14 +92,6 @@ jobs:
     with:
       language: >
         ["actions"]
-  joern-scan:
-    if: >
-      github.event_name == 'push'
-      || github.event_name == 'pull_request'
-      || (github.event_name == 'workflow_dispatch' && inputs.workflow == 'analyze')
-    permissions:
-      contents: read
-    uses: ./.github/workflows/joern-scan.yml
   github-release:
     if: >
       github.event_name == 'workflow_dispatch' && inputs.workflow == 'release'
--- a/.github/workflows/joern-scan.yml
+++ b/.github/workflows/joern-scan.yml
@@ -111,6 +111,22 @@ jobs:
         run: |
           [[ -n "${SCAN_PATH}" ]] || { echo '::error::scan-path must not be empty'; exit 1; }
 
+          # Canonicalize to an absolute path so scan-path can never be parsed as a
+          # `find`/joern-scan option (e.g. a caller-supplied "-delete"), and confirm
+          # it stays inside the checkout before it reaches `find` or joern-scan.
+          # GITHUB_WORKSPACE is resolved too since it may reach the workspace root
+          # through a symlink on self-hosted runners.
+          workspace_abs="$(realpath -e -- "${GITHUB_WORKSPACE}")"
+          scan_path_abs="$(realpath -e -- "${SCAN_PATH}")" || { echo '::error::scan-path does not exist'; exit 1; }
+          case "${scan_path_abs}" in
+            "${workspace_abs}" | "${workspace_abs}"/*) ;;
+            *)
+              echo '::error::scan-path must resolve to a location inside the workspace'
+              exit 1
+              ;;
+          esac
+          SCAN_PATH="${scan_path_abs}"
+
           # File-name patterns per Joern frontend, keyed by the frontend names joern-scan's
           # --language flag actually accepts, i.e. the `importCode.<name>` accessors in
           # joernio/joern's console/cpgcreation/ImportCode.scala. These differ from the
@@ -121,12 +137,12 @@ jobs:
           # upstream either (it targets arbitrary compiled binaries), so it is intentionally
           # left unvalidated here, same as an unrecognized language value.
           declare -A LANGUAGE_PATTERNS=(
-            [c]='*.c *.cc *.cpp *.h *.hpp *.hh'
-            [cpp]='*.c *.cc *.cpp *.h *.hpp *.hh'
+            [c]='*.c *.cc *.cpp *.cxx *.cp *.h *.hpp *.hh *.hxx *.h++ *.tcc'
+            [cpp]='*.c *.cc *.cpp *.cxx *.cp *.h *.hpp *.hh *.hxx *.h++ *.tcc'
             [java]='*.java'
             [jvm]='*.jar *.war *.ear *.apk *.class'
-            [jssrc]='*.js *.ts *.jsx *.tsx package.json'
-            [javascript]='*.js *.ts *.jsx *.tsx package.json'
+            [jssrc]='*.js *.ts *.jsx *.tsx *.cjs *.mjs *.vue *.ejs package.json'
+            [javascript]='*.js *.ts *.jsx *.tsx *.cjs *.mjs *.vue *.ejs package.json'
             [kotlin]='*.kt'
             [php]='*.php'
             [python]='*.py'
--- a/.github/workflows/joern-scan.yml
+++ b/.github/workflows/joern-scan.yml
@@ -1,0 +1,117 @@
+---
+name: Static analysis with Joern
+on:
+  workflow_call:
+    inputs:
+      scan-path:
+        required: false
+        type: string
+        description: Source code directory to scan
+        default: .
+      language:
+        required: false
+        type: string
+        description: Source language for joern-scan (auto-detected if not set)
+        default: null
+      joern-scan-options:
+        required: false
+        type: string
+        description: Additional space-separated command-line options for joern-scan
+        default: null
+      fail-on-findings:
+        required: false
+        type: boolean
+        description: Fail the job when joern-scan reports findings
+        default: true
+      java-version:
+        required: false
+        type: string
+        description: Java version to use
+        default: "21"
+      enable-cache:
+        required: false
+        type: boolean
+        description: Cache the pinned Joern download
+        default: true
+      cache-salt:
+        required: false
+        type: string
+        description: Optional value used to bust the Joern cache
+        default: null
+      runs-on:
+        required: false
+        type: string
+        description: Runner to use
+        default: ubuntu-latest
+permissions:
+  contents: read
+defaults:
+  run:
+    shell: bash -euo pipefail {0}
+    working-directory: .
+jobs:
+  joern-scan:
+    runs-on: ${{ inputs.runs-on }}
+    env:
+      # renovate: datasource=github-releases depName=joernio/joern
+      JOERN_VERSION: v4.0.579
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          fetch-depth: 1
+          persist-credentials: false
+      - name: Setup Java
+        uses: actions/setup-java@0f481fcb613427c0f801b606911222b5b6f3083a  # v5.5.0
+        with:
+          distribution: temurin
+          java-version: ${{ inputs.java-version }}
+      - name: Cache Joern
+        if: inputs.enable-cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+        with:
+          path: ~/.local/share/joern
+          key: joern-${{ runner.os }}-${{ runner.arch }}-${{ env.JOERN_VERSION }}-${{ inputs.cache-salt }}
+      - name: Install Joern and the default query database
+        env:
+          JOERN_VERSION: ${{ env.JOERN_VERSION }}
+        run: |
+          # shellcheck disable=SC2153
+          [[ "${JOERN_VERSION}" =~ ^v[0-9]+(\.[0-9]+)*([._+-][A-Za-z0-9]+)*$ ]] || { echo "invalid JOERN_VERSION"; exit 1; }
+          joern_home="${HOME}/.local/share/joern"
+          if [[ ! -x "${joern_home}/joern-cli/joern-scan" ]]; then
+            curl --fail --location --silent --show-error \
+              "https://github.com/joernio/joern/releases/download/${JOERN_VERSION}/joern-cli.zip" \
+              -o "${RUNNER_TEMP}/joern-cli.zip"
+            mkdir -p "${joern_home}"
+            unzip -qo "${RUNNER_TEMP}/joern-cli.zip" -d "${joern_home}"
+            # joern-scan prepends "v" to the version passed to --dbversion
+            "${joern_home}/joern-cli/joern-scan" --updatedb --dbversion "${JOERN_VERSION#v}"
+          fi
+          echo "${joern_home}/joern-cli" >> "${GITHUB_PATH}"
+      - name: Run joern-scan
+        env:
+          SCAN_PATH: ${{ inputs.scan-path }}
+          SCAN_LANGUAGE: ${{ inputs.language }}
+          JOERN_SCAN_OPTIONS: ${{ inputs.joern-scan-options }}
+          FAIL_ON_FINDINGS: ${{ inputs.fail-on-findings }}
+        run: |
+          [[ -n "${SCAN_PATH}" ]] || { echo '::error::scan-path must not be empty'; exit 1; }
+          args=()
+          if [[ -n "${SCAN_LANGUAGE}" ]]; then
+            args+=(--language "${SCAN_LANGUAGE}")
+          fi
+          if [[ -n "${JOERN_SCAN_OPTIONS}" ]]; then
+            read -r -a extra_args <<< "${JOERN_SCAN_OPTIONS}"
+            args+=("${extra_args[@]}")
+          fi
+          scan_log="${RUNNER_TEMP}/joern-scan.log"
+          joern-scan "${args[@]}" "${SCAN_PATH}" | tee "${scan_log}"
+          n_findings="$(grep -c '^Result: ' "${scan_log}" || true)"
+          if [[ "${n_findings}" -gt 0 ]]; then
+            if [[ "${FAIL_ON_FINDINGS}" == 'true' ]]; then
+              echo "::error::joern-scan reported ${n_findings} findings"
+              exit 1
+            fi
+            echo "::warning::joern-scan reported ${n_findings} findings"
+          fi
--- a/.github/workflows/joern-scan.yml
+++ b/.github/workflows/joern-scan.yml
@@ -16,7 +16,9 @@ on:
       joern-scan-options:
         required: false
         type: string
-        description: Additional space-separated command-line options for joern-scan
+        description: >-
+          Additional space-separated command-line options for joern-scan.
+          Limited to --names, --tags, --depth, --overwrite, and --store.
         default: null
       fail-on-findings:
         required: false
@@ -108,34 +110,97 @@ jobs:
           FAIL_ON_FINDINGS: ${{ inputs.fail-on-findings }}
         run: |
           [[ -n "${SCAN_PATH}" ]] || { echo '::error::scan-path must not be empty'; exit 1; }
+
+          # File-name patterns per Joern frontend, mirroring guessLanguageForRegularFile in
+          # joernio/joern's console/cpgcreation/package.scala. Keyed by the same identifiers
+          # accepted by joern-scan's --language flag. "ghidra" has no extension heuristic
+          # upstream either (it targets arbitrary compiled binaries), so it is intentionally
+          # left unvalidated here, same as an unrecognized language value.
+          declare -A LANGUAGE_PATTERNS=(
+            [c]='*.c *.cc *.cpp *.h *.hpp *.hh'
+            [newc]='*.c *.cc *.cpp *.h *.hpp *.hh'
+            [java]='*.jar *.war *.ear *.apk *.class'
+            [javasrc]='*.java'
+            [jssrc]='*.js *.ts *.jsx *.tsx package.json'
+            [javascript]='*.js *.ts *.jsx *.tsx package.json'
+            [kotlin]='*.kt'
+            [php]='*.php'
+            [pythonsrc]='*.py'
+            [python]='*.py'
+            [rubysrc]='*.rb'
+            [swiftsrc]='*.swift'
+            [golang]='*.go gopkg.lock gopkg.toml go.mod go.sum'
+            [csharp]='*.cs *.csproj'
+            [csharpsrc]='*.cs *.csproj'
+            [llvm]='*.bc *.ll'
+            [rust]='*.rs cargo.toml cargo.lock'
+            [abap]='*.abap'
+          )
+
+          # Checks whether scan-path contains a file matching any of the space-separated
+          # glob patterns in $1.
+          has_source_match() {
+            local -a name_args=()
+            local pattern
+            for pattern in $1; do
+              name_args+=(-o -iname "${pattern}")
+            done
+            [[ -n "$(find "${SCAN_PATH}" -type f \( "${name_args[@]:1}" \) -print -quit)" ]]
+          }
+
           args=()
           if [[ -n "${SCAN_LANGUAGE}" ]]; then
             args+=(--language "${SCAN_LANGUAGE}")
+            lang_key="${SCAN_LANGUAGE,,}"
+            lang_patterns="${LANGUAGE_PATTERNS[${lang_key}]:-}"
+            # Joern accepts an explicit --language without checking that scan-path
+            # actually contains matching source, so an unrelated or empty scan-path
+            # can still build an empty CPG and report a false-clean scan.
+            if [[ -n "${lang_patterns}" ]] && ! has_source_match "${lang_patterns}"; then
+              echo "::error::no source files matched by the '${SCAN_LANGUAGE}' frontend were found under scan-path"
+              exit 1
+            fi
           else
             # Joern silently falls back to the C frontend when auto-detection finds no
-            # match, producing an empty CPG and a false-clean scan. This mirrors
-            # guessLanguageForRegularFile in joernio/joern's console/cpgcreation/package.scala
-            # to reject scan paths with no file Joern can actually recognize.
-            if [[ -z "$(find "${SCAN_PATH}" -type f \( \
-                -iname '*.jar' -o -iname '*.war' -o -iname '*.ear' -o -iname '*.apk' \
-                -o -iname '*.java' -o -iname '*.class' -o -iname '*.kt' \
-                -o -iname '*.php' -o -iname '*.py' -o -iname '*.rb' -o -iname '*.swift' \
-                -o -iname '*.cs' -o -iname '*.csproj' \
-                -o -iname '*.go' -o -iname 'gopkg.lock' -o -iname 'gopkg.toml' -o -iname 'go.mod' -o -iname 'go.sum' \
-                -o -iname '*.bc' -o -iname '*.ll' \
-                -o -iname '*.js' -o -iname '*.ts' -o -iname '*.jsx' -o -iname '*.tsx' -o -iname 'package.json' \
-                -o -iname '*.c' -o -iname '*.cc' -o -iname '*.cpp' -o -iname '*.h' -o -iname '*.hpp' -o -iname '*.hh' \
-                -o -iname '*.rs' -o -iname 'cargo.toml' -o -iname 'cargo.lock' \
-                -o -iname '*.abap' \
-              \) -print -quit)" ]]; then
+            # match, producing an empty CPG and a false-clean scan.
+            if ! has_source_match "${LANGUAGE_PATTERNS[*]}"; then
               echo '::error::no source files matched by a known Joern frontend were found under scan-path; set the language input explicitly or verify scan-path'
               exit 1
             fi
           fi
+
           if [[ -n "${JOERN_SCAN_OPTIONS}" ]]; then
             read -r -a extra_args <<< "${JOERN_SCAN_OPTIONS}"
+            # Allow-list only scan-tuning flags. Anything else -- non-scan modes such as
+            # --dump/--list-query-names/--list-languages/--updatedb, --language (has its
+            # own input), --frontend-args, --help, or a bare positional path -- can make
+            # joern-scan exit 0 without analyzing scan-path at all, which the findings
+            # check below cannot detect.
+            expect_value=0
+            for extra_arg in "${extra_args[@]}"; do
+              if [[ "${expect_value}" -eq 1 ]]; then
+                expect_value=0
+                continue
+              fi
+              case "${extra_arg}" in
+                --names | --tags | --depth)
+                  expect_value=1
+                  ;;
+                --overwrite | --store)
+                  ;;
+                *)
+                  echo "::error::joern-scan-options contains an unsupported option '${extra_arg}'; only --names, --tags, --depth, --overwrite, and --store are allowed"
+                  exit 1
+                  ;;
+              esac
+            done
+            if [[ "${expect_value}" -eq 1 ]]; then
+              echo '::error::joern-scan-options ends with an option that requires a value'
+              exit 1
+            fi
             args+=("${extra_args[@]}")
           fi
+
           scan_log="${RUNNER_TEMP}/joern-scan.log"
           joern-scan "${args[@]}" -- "${SCAN_PATH}" 2>&1 | tee "${scan_log}"
           # "Error executing query" is logged via logger.warn (may carry a
@@ -143,6 +208,13 @@ jobs:
           # line start.
           if grep -q -E 'Error executing query|No queries matched current filter selection' "${scan_log}"; then
             echo '::error::joern-scan did not complete a full analysis (see log for query execution errors)'
+            exit 1
+          fi
+          # joern-scan only prints this line after a scan runs to completion
+          # (JoernScan.runScanPlugin's final println); a bypassed or aborted run
+          # leaves it out even with no error and zero findings.
+          if ! grep -q -F 'to explore interactively' "${scan_log}"; then
+            echo '::error::joern-scan did not report a completed scan (no completion message found in output)'
             exit 1
           fi
           n_findings="$(grep -c '^Result: ' "${scan_log}" || true)"
--- a/.github/workflows/joern-scan.yml
+++ b/.github/workflows/joern-scan.yml
@@ -66,11 +66,12 @@ jobs:
         with:
           distribution: temurin
           java-version: ${{ inputs.java-version }}
-      - name: Cache Joern
+      - name: Restore Joern cache
         if: inputs.enable-cache
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+        id: joern-cache
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
         with:
-          path: ~/.local/share/joern
+          path: ~/.local/share/joern/${{ env.JOERN_VERSION }}
           key: joern-${{ runner.os }}-${{ runner.arch }}-${{ env.JOERN_VERSION }}-${{ inputs.cache-salt }}
       - name: Install Joern and the default query database
         env:
@@ -78,7 +79,9 @@ jobs:
         run: |
           # shellcheck disable=SC2153
           [[ "${JOERN_VERSION}" =~ ^v[0-9]+(\.[0-9]+)*([._+-][A-Za-z0-9]+)*$ ]] || { echo "invalid JOERN_VERSION"; exit 1; }
-          joern_home="${HOME}/.local/share/joern"
+          # Version-qualified so a stale install from a different JOERN_VERSION on a
+          # persistent runner can never satisfy this check after a cache miss.
+          joern_home="${HOME}/.local/share/joern/${JOERN_VERSION}"
           if [[ ! -x "${joern_home}/joern-cli/joern-scan" ]]; then
             curl --fail --location --silent --show-error \
               --connect-timeout 30 --speed-limit 51200 --speed-time 60 \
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -92,6 +92,14 @@ jobs:
...
       language: >
         ["actions"]
+  joern-scan:
+    if: >
+      github.event_name == 'push'
+      || github.event_name == 'pull_request'
+      || (github.event_name == 'workflow_dispatch' && inputs.workflow == 'analyze')
+    permissions:
+      contents: read
+    uses: ./.github/workflows/joern-scan.yml
```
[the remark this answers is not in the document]
dceoy: Agreed — removed the joern-scan job from ci.yml in b0d9a84 since this repo has no Joern-supported source, so this integration could never produce meaningful results.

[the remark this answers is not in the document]
dceoy: Confirmed via the PR's CI run: the job passes but silently falls back to the "c" language with an empty CPG, so it always reports 0 findings. Removed the joern-scan job from ci.yml in b0d9a84; the reusable joern-scan.yml workflow itself is unaffected.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+          joern_home="${HOME}/.local/share/joern"
+          if [[ ! -x "${joern_home}/joern-cli/joern-scan" ]]; then
+            curl --fail --location --silent --show-error \
+              "https://github.com/joernio/joern/releases/download/${JOERN_VERSION}/joern-cli.zip" \
+              -o "${RUNNER_TEMP}/joern-cli.zip"
+            mkdir -p "${joern_home}"
+            unzip -qo "${RUNNER_TEMP}/joern-cli.zip" -d "${joern_home}"
+            # joern-scan prepends "v" to the version passed to --dbversion
+            "${joern_home}/joern-cli/joern-scan" --updatedb --dbversion "${JOERN_VERSION#v}"
+          fi
```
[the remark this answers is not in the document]
dceoy: Agreed this is a residual gap, but leaving as-is for now since upstream Joern releases don't publish a checksum/signature to verify against.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+    working-directory: .
+jobs:
+  joern-scan:
+    runs-on: ${{ inputs.runs-on }}
+    env:
+      # renovate: datasource=github-releases depName=joernio/joern
+      JOERN_VERSION: v4.0.579
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
```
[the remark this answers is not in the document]
dceoy: Fixed in b0d9a84 — now pins the same v7.0.1 SHA used elsewhere in the repo.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+          key: joern-${{ runner.os }}-${{ runner.arch }}-${{ env.JOERN_VERSION }}-${{ inputs.cache-salt }}
+      - name: Install Joern and the default query database
+        env:
+          JOERN_VERSION: ${{ env.JOERN_VERSION }}
+        run: |
+          # shellcheck disable=SC2153
+          [[ "${JOERN_VERSION}" =~ ^v[0-9]+(\.[0-9]+)*([._+-][A-Za-z0-9]+)*$ ]] || { echo "invalid JOERN_VERSION"; exit 1; }
+          joern_home="${HOME}/.local/share/joern"
+          if [[ ! -x "${joern_home}/joern-cli/joern-scan" ]]; then
+            curl --fail --location --silent --show-error \
```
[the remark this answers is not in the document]
dceoy: Fixed in b0d9a84 — added --connect-timeout plus --speed-limit/--speed-time to fail fast on a stalled connection without capping total transfer time on a slow link.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+              "https://github.com/joernio/joern/releases/download/${JOERN_VERSION}/joern-cli.zip" \
+              -o "${RUNNER_TEMP}/joern-cli.zip"
+            mkdir -p "${joern_home}"
+            unzip -qo "${RUNNER_TEMP}/joern-cli.zip" -d "${joern_home}"
```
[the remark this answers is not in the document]
dceoy: Added an rm -rf of the old joern-cli dir before unzip in b0d9a84 as a cheap safeguard, though note the existing -x guard already skips this whole block (including the unzip) once a joern-scan binary exists, so it isn't reached on a version bump without a cache miss.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+          args=()
+          if [[ -n "${SCAN_LANGUAGE}" ]]; then
+            args+=(--language "${SCAN_LANGUAGE}")
+          fi
+          if [[ -n "${JOERN_SCAN_OPTIONS}" ]]; then
+            read -r -a extra_args <<< "${JOERN_SCAN_OPTIONS}"
+            args+=("${extra_args[@]}")
+          fi
+          scan_log="${RUNNER_TEMP}/joern-scan.log"
+          joern-scan "${args[@]}" "${SCAN_PATH}" | tee "${scan_log}"
```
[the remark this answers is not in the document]
dceoy: Fixed in b0d9a84 — added a `--` separator before the scan path; joern-scan uses scopt, which honors it as an end-of-options marker.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,119 @@
...
+          if [[ -n "${JOERN_SCAN_OPTIONS}" ]]; then
+            read -r -a extra_args <<< "${JOERN_SCAN_OPTIONS}"
+            args+=("${extra_args[@]}")
+          fi
+          scan_log="${RUNNER_TEMP}/joern-scan.log"
+          joern-scan "${args[@]}" -- "${SCAN_PATH}" | tee "${scan_log}"
+          n_findings="$(grep -c '^Result: ' "${scan_log}" || true)"
+          if [[ "${n_findings}" -gt 0 ]]; then
+            if [[ "${FAIL_ON_FINDINGS}" == 'true' ]]; then
+              echo "::error::joern-scan reported ${n_findings} findings"
```
dceoy, line 333: **P1 · Fail when no supported language can be detected**

Leaving `language` unset is described as automatic detection, but Joern falls back to the C frontend when detection returns no match. A caller scanning only Shell, YAML, Markdown, or another unsupported source set therefore builds an empty C CPG and reports zero findings, creating a false-clean result.

Either require `language`, or add a preflight that rejects inputs containing no files supported by the selected/detected frontend.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,119 @@
...
+              exit 1
```
dceoy, line 334: **P1 · Do not treat query execution failures as a clean scan**

Counting only `^Result: ` lines cannot distinguish “no vulnerabilities” from incomplete analysis. Joern catches exceptions thrown by individual queries, logs `Error executing query`, and returns an empty result; it also returns normally when filters match no queries. Both cases currently leave `n_findings=0` and the workflow green.

Capture stderr as well as stdout and fail on known analysis-failure diagnostics such as `Error executing query` and `No queries matched current filter selection`, or consume a machine-readable status if Joern adds one.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,119 @@
...
+          JOERN_VERSION: ${{ env.JOERN_VERSION }}
+        run: |
+          # shellcheck disable=SC2153
+          [[ "${JOERN_VERSION}" =~ ^v[0-9]+(\.[0-9]+)*([._+-][A-Za-z0-9]+)*$ ]] || { echo "invalid JOERN_VERSION"; exit 1; }
+          joern_home="${HOME}/.local/share/joern"
+          if [[ ! -x "${joern_home}/joern-cli/joern-scan" ]]; then
+            curl --fail --location --silent --show-error \
+              --connect-timeout 30 --speed-limit 51200 --speed-time 60 \
+              "https://github.com/joernio/joern/releases/download/${JOERN_VERSION}/joern-cli.zip" \
+              -o "${RUNNER_TEMP}/joern-cli.zip"
```
dceoy, line 115: **P2 · Enforce the pinned Joern version on persistent runners**

The installation directory is shared by every `JOERN_VERSION`, and this guard accepts any existing `joern-scan` executable. On a self-hosted runner, after the pinned version changes and the new cache key misses, an older installation can satisfy this check and continue running indefinitely. The `rm -rf` below is unreachable in that case.

Use a versioned path such as `${HOME}/.local/share/joern/${JOERN_VERSION}`, or persist and validate an installed-version marker before skipping installation.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,119 @@
...
+        with:
+          distribution: temurin
+          java-version: ${{ inputs.java-version }}
+      - name: Cache Joern
+        if: inputs.enable-cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+        with:
+          path: ~/.local/share/joern
+          key: joern-${{ runner.os }}-${{ runner.arch }}-${{ env.JOERN_VERSION }}-${{ inputs.cache-salt }}
+      - name: Install Joern and the default query database
```
dceoy, line 103: **P2 · Save the installation before findings can fail the job**

`actions/cache` saves in its post-step only when the job succeeds. Since `fail-on-findings` defaults to `true`, the first scan of a vulnerable repository exits after installation and prevents the Joern cache from being written. Every subsequent run then repeats the CLI and QueryDB downloads until the code becomes clean.

Use separate `actions/cache/restore` and `actions/cache/save` steps, and save immediately after successful installation before running the scan.

dceoy (commented):
The workflow is structurally clean and the prior repository-level invocation issue has been removed, but four correctness/operability issues remain. Two can produce false-clean scans, so these should be addressed before merge.

.github/workflows/joern-scan.yml
dceoy: Fixed in 2c283232d11c01ffe0016a09bfe85e521c71a6c2 — added a preflight check that fails when no file under scan-path matches a known Joern frontend extension (mirroring guessLanguageForRegularFile), instead of silently falling back to C. Note: this only guards the auto-detect (unset language) case described here; passing an explicit `language` that doesn't match any files in scan-path can still produce the same false-clean empty-CPG result — flagging as a possible follow-up.

.github/workflows/joern-scan.yml
dceoy: Fixed in 2c283232d11c01ffe0016a09bfe85e521c71a6c2 — joern-scan's combined stdout+stderr output is now checked for "Error executing query" (logged via logger.warn) and "No queries matched current filter selection" (println), failing the job if either appears instead of reporting 0 findings.

.github/workflows/joern-scan.yml
dceoy: Fixed in 2c283232d11c01ffe0016a09bfe85e521c71a6c2 — the install path is now version-qualified (`~/.local/share/joern/$JOERN_VERSION`), so a stale install from a different version can no longer satisfy the skip-install check. Trade-off: on persistent runners, obsolete version directories now accumulate instead of obsolete files inside one directory, since this workflow doesn't prune old versions.

.github/workflows/joern-scan.yml
dceoy: Fixed in 2c283232d11c01ffe0016a09bfe85e521c71a6c2 — split the single cache step into `actions/cache/restore` (before install) and `actions/cache/save` (right after install succeeds, before the scan step that can fail the job on findings).

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,155 @@
...
+                -o -iname '*.js' -o -iname '*.ts' -o -iname '*.jsx' -o -iname '*.tsx' -o -iname 'package.json' \
+                -o -iname '*.c' -o -iname '*.cc' -o -iname '*.cpp' -o -iname '*.h' -o -iname '*.hpp' -o -iname '*.hh' \
+                -o -iname '*.rs' -o -iname 'cargo.toml' -o -iname 'cargo.lock' \
+                -o -iname '*.abap' \
+              \) -print -quit)" ]]; then
+              echo '::error::no source files matched by a known Joern frontend were found under scan-path; set the language input explicitly or verify scan-path'
+              exit 1
+            fi
+          fi
+          if [[ -n "${JOERN_SCAN_OPTIONS}" ]]; then
```
dceoy, line 270: **P1 — Restrict options that bypass the scan**

`joern-scan-options` is appended verbatim to the scanner arguments, but Joern accepts non-scan modes such as `--dump`, `--list-query-names`, `--list-languages`, and `--updatedb`. Those modes return without calling the scan path, and the subsequent checks see no error diagnostic and zero `Result:` lines, so the job succeeds without analyzing the repository.

`--frontend-args` is another bypass: Joern splits all following arguments into frontend arguments. Because this workflow appends the source path last, a caller can move the path out of the scanner arguments; Joern then prints usage for an empty source and returns normally.

Prefer typed inputs for supported scan controls (`names`, `tags`, `depth`, and optionally dedicated frontend arguments). At minimum, reject non-scan modes, `--language`, `--frontend-args`, and extra positional arguments, and verify a positive scan-completion signal rather than inferring completion only from the absence of known errors.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,155 @@
...
+          key: ${{ steps.joern-cache.outputs.cache-primary-key }}
+      - name: Run joern-scan
+        env:
+          SCAN_PATH: ${{ inputs.scan-path }}
+          SCAN_LANGUAGE: ${{ inputs.language }}
+          JOERN_SCAN_OPTIONS: ${{ inputs.joern-scan-options }}
+          FAIL_ON_FINDINGS: ${{ inputs.fail-on-findings }}
+        run: |
+          [[ -n "${SCAN_PATH}" ]] || { echo '::error::scan-path must not be empty'; exit 1; }
+          args=()
```
dceoy: **P1 — Validate explicitly selected languages against the source tree**

The no-supported-source preflight only runs when `language` is unset. With an explicit value, the workflow forwards `--language` without checking that the scan path contains any source handled by that frontend. For example, a Python-only directory with `language: c` can produce an empty or irrelevant CPG and still be reported as a clean scan.

Apply the same source-presence validation for explicit languages, using the extension/file-name set for the selected frontend. Also prevent `joern-scan-options` from supplying a second `--language` that bypasses the dedicated input.

dceoy (commented):
Two false-clean paths remain in the reusable workflow. Both can make a run succeed with zero findings even though the requested source analysis was not performed. Details are in the inline comments.

.github/workflows/joern-scan.yml
dceoy: Fixed in fd17d59 — joern-scan-options is now allow-listed to --names/--tags/--depth/--overwrite/--store, and the run is verified for joern-scan's completion message rather than just the absence of errors.

.github/workflows/joern-scan.yml
dceoy: Fixed in fd17d59 — the source-presence preflight now also runs for an explicit language input, checked against that frontend's own file patterns, and joern-scan-options can no longer supply a second --language.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,227 @@
...
+          # joernio/joern's console/cpgcreation/package.scala. Keyed by the same identifiers
+          # accepted by joern-scan's --language flag. "ghidra" has no extension heuristic
+          # upstream either (it targets arbitrary compiled binaries), so it is intentionally
+          # left unvalidated here, same as an unrecognized language value.
+          declare -A LANGUAGE_PATTERNS=(
+            [c]='*.c *.cc *.cpp *.h *.hpp *.hh'
+            [newc]='*.c *.cc *.cpp *.h *.hpp *.hh'
+            [java]='*.jar *.war *.ear *.apk *.class'
+            [javasrc]='*.java'
+            [jssrc]='*.js *.ts *.jsx *.tsx package.json'
```
dceoy: **P1 — Align language validation with Joern's actual frontend names**

These keys mix Joern's generated language identifiers with the `importCode.<frontend>` names that `joern-scan --language` actually invokes. In Joern v4.0.579, `language: java` selects the Java **source** frontend, while JVM bytecode uses `jvm`; the available frontend names also include `cpp`, `python`, and `ruby`, whereas aliases such as `javasrc`, `pythonsrc`, `rubysrc`, and `newc` are not `importCode` methods.

Consequently, a normal `.java` repository with `language: java` is rejected here because `java` is mapped only to archives/classes; `language: javasrc` passes this preflight but later fails when Joern tries to invoke a nonexistent frontend; and valid values such as `jvm`, `cpp`, and `ruby` bypass source-presence validation because they are absent from this map.

Please key the map by the actual frontend names accepted by the scanner, for example `java -> *.java`, `jvm -> *.jar *.war *.ear *.apk *.class`, `c/cpp -> C/C++ patterns`, `python -> *.py`, and `ruby -> *.rb`, or normalize aliases to those canonical names before both validation and invocation.

dceoy (commented):
One blocking correctness issue remains in the explicit-language validation. The preflight language keys need to match the frontend method names that Joern actually invokes; otherwise valid configurations are rejected and invalid aliases pass the preflight only to fail later.

.github/workflows/joern-scan.yml
dceoy: Fixed in 12d6834 — re-keyed the map to joern-scan's actual `importCode.<name>` frontend names (verified against joernio/joern's `ImportCode.scala` at the pinned v4.0.579 tag): `java` now validates `*.java` source, `jvm` was added for bytecode, and `cpp`/`ruby` were added while `javasrc`/`pythonsrc`/`rubysrc`/`newc` (not real frontend names) were dropped.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,230 @@
...
+          # glob patterns in $1.
+          has_source_match() {
+            local -a name_args=()
+            local pattern
+            for pattern in $1; do
+              name_args+=(-o -iname "${pattern}")
+            done
+            [[ -n "$(find "${SCAN_PATH}" -type f \( "${name_args[@]:1}" \) -print -quit)" ]]
+          }
+
```
dceoy, line 212: **P1 — Prevent `scan-path` from being parsed as a `find` expression**

`SCAN_PATH` is caller-controlled and is passed as the first argument to GNU `find`. Quoting does not make a leading `-` safe: for example, `scan-path: -delete` results in `find -delete -type f ...`, which defaults to the current directory and executes the destructive `-delete` action. I reproduced this behavior locally with GNU find.

This can modify the checked-out workspace and is more consequential on persistent self-hosted runners allowed by `runs-on`.

Canonicalize the path with `realpath --`, verify that it is inside `GITHUB_WORKSPACE`, and pass only the validated absolute path to both `find` and Joern. At minimum, use an explicit end-of-options mechanism supported by the target `find`, but workspace-boundary validation is preferable.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,230 @@
...
+            [ruby]='*.rb'
+            [swiftsrc]='*.swift'
+            [golang]='*.go gopkg.lock gopkg.toml go.mod go.sum'
+            [csharp]='*.cs *.csproj'
+            [csharpsrc]='*.cs *.csproj'
+            [llvm]='*.bc *.ll'
+            [rust]='*.rs cargo.toml cargo.lock'
+            [abap]='*.abap'
+          )
+
```
dceoy: **P2 — Do not reject source extensions supported by the selected Joern frontend**

This map is suitable only as a simplified detection heuristic; it is narrower than the files the actual frontends accept. For example, Joern's C/C++ frontend supports extensions including `.cxx`, `.cp`, `.hxx`, `.h++`, and `.tcc`, while the JavaScript frontend also handles `.cjs`, `.mjs`, `.vue`, and `.ejs`. A valid project containing only one of those extensions fails this preflight even though Joern can process it.

Separate automatic-language detection patterns from explicit-frontend validation, and make the latter reflect each frontend's real supported source extensions. Add regression cases such as `language: cpp` with a `.cxx` file and `language: jssrc` with an `.mjs` file.

dceoy (commented):
Two issues remain before merge:

1. `scan-path` is passed to GNU `find` as its first argument without an end-of-options safeguard or workspace-boundary validation. A value such as `-delete` is parsed as a `find` expression and can delete files from the current working directory, which is particularly risky on persistent self-hosted runners.
2. The explicit-language preflight patterns are narrower than the actual Joern frontend support, causing valid projects such as `.cxx`-only C++ or `.mjs`-only JavaScript trees to be rejected before Joern runs.

The previous review threads are resolved and the current CI run is green, but this repository no longer invokes the new reusable workflow itself, so these runtime paths are not covered by CI. Please fix both issues and add focused regression tests for hostile `scan-path` values and supported extensions omitted from the current map.

.github/workflows/joern-scan.yml
dceoy: Fixed in 4b57f8d: scan-path is now canonicalized with `realpath -e --` (both sides resolved, so a symlinked workspace root on self-hosted runners still matches) and rejected unless it resolves inside GITHUB_WORKSPACE, before it reaches `find` or joern-scan. Went with the workspace-boundary validation option as suggested.

.github/workflows/joern-scan.yml
dceoy: Fixed in 4b57f8d: widened the c/cpp and jssrc/javascript patterns to include .cxx/.cp/.hxx/.h++/.tcc and .cjs/.mjs/.vue/.ejs. Kept a single shared map rather than splitting auto-detect from explicit-frontend validation — the auto-detect branch only does a presence check over all patterns, so a broader list cannot misroute detection, and a second map would just duplicate it. Did not add fixture-based regression cases since this repo has no test harness for reusable-workflow shell logic (the only bats suite covers the README generator).

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,246 @@
...
+              echo '::error::joern-scan-options ends with an option that requires a value'
+              exit 1
+            fi
+            args+=("${extra_args[@]}")
+          fi
+
+          scan_log="${RUNNER_TEMP}/joern-scan.log"
+          joern-scan "${args[@]}" -- "${SCAN_PATH}" 2>&1 | tee "${scan_log}"
+          # "Error executing query" is logged via logger.warn (may carry a
+          # timestamp/level prefix); match as a substring rather than anchoring to
```
dceoy, line 317: **P1 — Rebuild the CPG on every CI run**

`joern-scan` defaults to reusing an existing workspace project unless `--overwrite` is set. Joern creates that workspace under `$INSTALL_DIR/workspace`; here `$INSTALL_DIR` is the versioned `~/.local/share/joern/.../joern-cli` tree. On a persistent self-hosted runner allowed by `runs-on`, the first scan therefore leaves a CPG keyed by the stable checkout path, and subsequent runs print `Using existing CPG` and analyze that old graph even after the repository changes. The completion-message and finding checks still pass, producing a false-clean result for the current commit.

Pass `--overwrite` unconditionally for CI scans (rather than leaving it as an optional caller flag), or remove the matching workspace project before invoking Joern.

dceoy (commented):
One blocking correctness issue remains for the supported self-hosted-runner path. Joern persists its workspace under the installation directory, and the current invocation permits reusing an existing CPG, so later runs can scan stale code instead of the current checkout.

.github/workflows/joern-scan.yml
dceoy: Fixed in 04f3560 — `--overwrite` is now passed unconditionally and removed from the `joern-scan-options` allow-list.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,249 @@
...
+          }
+
+          # --overwrite forces a fresh CPG for every run; joern-scan otherwise reuses
+          # any existing workspace project keyed by the checkout path, which on a
+          # persistent self-hosted runner would silently re-analyze a stale CPG.
+          args=(--overwrite)
+          if [[ -n "${SCAN_LANGUAGE}" ]]; then
+            args+=(--language "${SCAN_LANGUAGE}")
+            lang_key="${SCAN_LANGUAGE,,}"
+            lang_patterns="${LANGUAGE_PATTERNS[${lang_key}]:-}"
```
dceoy: **P1 — Validate values before Joern embeds them into generated Scala**

An unrecognized `language` is forwarded unchanged, and `--names`/`--tags` values are only tokenized. Joern v4.0.579 interpolates these values into a generated Scala script (`importCode.$language(...)` and `Array("$value")`) without safely escaping them. Quotes or Scala syntax in an input can therefore break the scan or execute injected Scala/Java code on the runner.

Reject languages outside the supported frontend allow-list and normalize them before invocation. Prefer dedicated typed inputs for names/tags/depth/store, with strict character validation for names and tags, instead of forwarding a generic option string.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,249 @@
...
+            # actually contains matching source, so an unrelated or empty scan-path
+            # can still build an empty CPG and report a false-clean scan.
+            if [[ -n "${lang_patterns}" ]] && ! has_source_match "${lang_patterns}"; then
+              echo "::error::no source files matched by the '${SCAN_LANGUAGE}' frontend were found under scan-path"
+              exit 1
+            fi
+          else
+            # Joern silently falls back to the C frontend when auto-detection finds no
+            # match, producing an empty CPG and a false-clean scan.
+            if ! has_source_match "${LANGUAGE_PATTERNS[*]}"; then
```
dceoy: **P1 — Keep the auto-detection preflight identical to Joern's detector**

This preflight uses the broader explicit-frontend patterns. For example, a repository containing only `.mjs`, `.cjs`, `.vue`, or `.ejs` passes this check, but Joern v4.0.579's `guessLanguageForRegularFile` does not recognize those extensions. Joern then falls back to the C frontend, builds an empty CPG, and reports a false-clean zero-finding scan.

Use a separate auto-detection pattern set that exactly mirrors Joern's detector. Keep the broader patterns only for an explicitly selected frontend, so files supported by a frontend but not auto-detected require `language` to be set.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,249 @@
...
+            [rust]='*.rs cargo.toml cargo.lock'
+            [abap]='*.abap'
+          )
+
+          # Checks whether scan-path contains a file matching any of the space-separated
+          # glob patterns in $1.
+          has_source_match() {
+            local -a name_args=()
+            local pattern
+            for pattern in $1; do
```
dceoy: **P2 — Prevent shell pathname expansion of validation patterns**

`for pattern in $1` performs both word splitting and pathname expansion in the workflow's current directory. If a pattern such as `*.py` matches a file at the repository root, the loop receives that concrete filename rather than the glob. A different `scan-path` containing valid Python sources can then be rejected incorrectly.

Parse the pattern string into an array first and iterate over quoted elements, for example:

```bash
local -a patterns=()
read -r -a patterns <<< "$1"
for pattern in "${patterns[@]}"; do
  name_args+=(-o -iname "$pattern")
done
```

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,249 @@
...
+      - name: Install Joern and the default query database
+        env:
+          JOERN_VERSION: ${{ env.JOERN_VERSION }}
+        run: |
+          # shellcheck disable=SC2153
+          [[ "${JOERN_VERSION}" =~ ^v[0-9]+(\.[0-9]+)*([._+-][A-Za-z0-9]+)*$ ]] || { echo "invalid JOERN_VERSION"; exit 1; }
+          # Version-qualified so a stale install from a different JOERN_VERSION on a
+          # persistent runner can never satisfy this check after a cache miss.
+          joern_home="${HOME}/.local/share/joern/${JOERN_VERSION}"
+          if [[ ! -x "${joern_home}/joern-cli/joern-scan" ]]; then
```
dceoy: **P2 — Mark the complete installation, not only the CLI executable**

If extraction succeeds but `joern-scan --updatedb` fails, a persistent runner retains an executable and this guard skips installation on the next run. With no query database, Joern downloads its default `latest` database during the scan and exits, so retries can fail once and then use a query bundle that is no longer tied to `JOERN_VERSION`.

Create a completion marker only after `--updatedb` succeeds and test that marker here, or build the full installation in a temporary directory and atomically rename it into place.

dceoy (commented):
The latest implementation still has two blocking false-clean/security issues and two robustness issues. See the inline findings below.

.github/workflows/joern-scan.yml
dceoy: Fixed in 260315e — `language` is now rejected unless it matches the frontend allow-list (or is `ghidra`), only the validated/normalized key is forwarded to `--language`, and `--names`/`--tags`/`--depth` values in `joern-scan-options` are restricted to safe characters — joern-scan interpolates all of these unescaped into the generated Scala script (verified against `BridgeBase.loadOrCreateCpg` at the pinned tag).

.github/workflows/joern-scan.yml
dceoy: Fixed in 260315e, reversing my position in 4b57f8d: auto-detection now uses its own narrower pattern set mirroring `guessLanguageForRegularFile` (no .cjs/.mjs/.vue/.ejs or the widened C/C++ extensions), since those extra patterns pass this preflight but Joern's real detector doesn't recognize them.

.github/workflows/joern-scan.yml
dceoy: Fixed in 260315e — `has_source_match` now parses the pattern string into an array via `read -a` before iterating, instead of word-splitting $1 directly, so a matching file in the workflow's cwd can no longer override a literal glob pattern.

.github/workflows/joern-scan.yml
dceoy: Fixed in 260315e — installation is now marked complete via a marker file (inside the cached `joern_home`) written only after both `joern-cli` extraction and `--updatedb` succeed, instead of just checking for the `joern-scan` executable.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,289 @@
...
+  contents: read
+defaults:
+  run:
+    shell: bash -euo pipefail {0}
+    working-directory: .
+jobs:
+  joern-scan:
+    runs-on: ${{ inputs.runs-on }}
+    env:
+      # renovate: datasource=github-releases depName=joernio/joern
```
dceoy, line 60: **HIGH (P1) — `runs-on` permits runners that cannot execute this script**

This input accepts arbitrary runner labels, but the scan step requires Bash 4+ features: associative arrays (`declare -A`) and lowercase expansion (`${SCAN_LANGUAGE,,}`). GitHub-hosted macOS runners still provide Bash 3.2, so `runs-on: macos-latest` fails before Joern starts even though the workflow exposes it as a supported configuration.

Either constrain/document this input as Linux-only, or install and explicitly invoke a Bash version that supports the script on every allowed runner. Confidence: HIGH.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,289 @@
...
+
+          # File-name patterns Joern's own auto-detection heuristic recognizes
+          # (guessLanguageForRegularFile in console/cpgcreation/package.scala at the
+          # pinned JOERN_VERSION). This is intentionally narrower than LANGUAGE_PATTERNS
+          # above -- e.g. it has no .cjs/.mjs/.vue/.ejs or .cxx/.hxx/.tcc entries -- so
+          # auto-detection here must not pass on a file Joern's real detector would not
+          # recognize, which would fall back to the C frontend and build an empty CPG.
+          AUTO_DETECT_PATTERNS='*.jar *.war *.ear *.apk *.csproj *.cs *.go gopkg.lock gopkg.toml go.mod go.sum *.js *.ts *.jsx *.tsx package.json *.java *.class *.kt *.php *.py *.rb *.swift *.bc *.ll *.c *.cc *.cpp *.h *.hpp *.hh *.rs cargo.toml cargo.lock *.abap'
+
+          # Checks whether scan-path contains a file matching any of the space-separated
```
dceoy: **HIGH (P1) — Require actual source files, not only language-detection manifests**

This preflight treats files such as `package.json`, `go.mod`, `go.sum`, `cargo.toml`, and `cargo.lock` as sufficient proof that analyzable source exists. Those files are useful for Joern's language heuristic, but a directory containing only a manifest can still produce an empty or near-empty CPG and complete with zero findings. That preserves a false-clean path when `scan-path` points at the wrong subdirectory.

Separate language-detection hints from source-presence validation. After choosing a frontend, require at least one real source file (or an expected analyzable binary for binary frontends) before reporting a successful scan. Confidence: HIGH.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,289 @@
...
+          java-version: ${{ inputs.java-version }}
+      - name: Restore Joern cache
+        if: inputs.enable-cache
+        id: joern-cache
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+        with:
+          path: ~/.local/share/joern/${{ env.JOERN_VERSION }}
+          key: joern-${{ runner.os }}-${{ runner.arch }}-${{ env.JOERN_VERSION }}-${{ inputs.cache-salt }}
+      - name: Install Joern and the default query database
+        env:
```
dceoy: **MEDIUM (P2) — `cache-salt` does not invalidate a persistent local installation**

The salt changes the cache key, but `joern_home` and `.install-complete` are keyed only by `JOERN_VERSION`. On a persistent self-hosted runner, a cache miss leaves the existing local directory intact; the marker then skips installation, and the same files are saved under the new salted key. Consequently, changing `cache-salt` cannot recover from a corrupted/stale local installation or query database.

Include the salt (or a hash of the full cache key) in `joern_home`, or record it in the completion marker and remove/reinstall when it differs. Confidence: HIGH.

dceoy (commented):
## Important Issues

Posted 3 inline findings against the current head: two HIGH-severity correctness/compatibility risks and one MEDIUM-severity cache invalidation issue.

## Checked

- Current head: `260315e7fb993c210de192fb0d8216bfd1071af4`
- Existing review threads and their resolutions
- Latest successful CI run
- Joern v4.0.579 implementation details relevant to language detection and scan completion

## Recommended Action

Address the two HIGH-severity findings before merge. The cache-salt behavior should also be corrected if persistent self-hosted runners are an intended use case.

.github/workflows/joern-scan.yml
dceoy: Fixed in 8ff589f: added a preflight step that fails fast with a clear error when `runner.os` isn't Linux, and documented the Linux-only requirement on the `runs-on` input.

.github/workflows/joern-scan.yml
dceoy: Fixed in 8ff589f: dropped manifest/lockfiles (package.json, go.mod, go.sum, cargo.toml, cargo.lock, gopkg.*) from the language and auto-detect source-presence patterns, so a directory containing only a manifest no longer passes the pre-scan check.

.github/workflows/joern-scan.yml
dceoy: Fixed in 8ff589f: the install directory is now salted with a hash of `cache-salt` (shared by cache restore/save and the install step), so bumping the salt on a persistent runner starts from a clean directory instead of reusing the untouched local install.

.github/workflows/joern-scan.yml
```diff
@@ -0,0 +1,315 @@
...
+          # (guessLanguageForRegularFile in console/cpgcreation/package.scala at the
+          # pinned JOERN_VERSION), minus the manifest/lockfiles excluded above for the
+          # same false-clean reason. This is intentionally narrower than
+          # LANGUAGE_PATTERNS above -- e.g. it has no .cjs/.mjs/.vue/.ejs or
+          # .cxx/.hxx/.tcc entries -- so auto-detection here must not pass on a file
+          # Joern's real detector would not recognize, which would fall back to the C
+          # frontend and build an empty CPG.
+          AUTO_DETECT_PATTERNS='*.jar *.war *.ear *.apk *.csproj *.cs *.go *.js *.ts *.jsx *.tsx *.java *.class *.kt *.php *.py *.rb *.swift *.bc *.ll *.c *.cc *.cpp *.h *.hpp *.hh *.rs *.abap'
+
+          # Checks whether scan-path contains a file matching any of the space-separated
```
dceoy: **P1 — Validate source files for the frontend Joern will actually select**

This preflight only proves that *some* supported source file exists. Joern's auto-detection instead counts all recognized files and selects the majority language, including manifest files such as `package.json`, `go.mod`, `go.sum`, `cargo.toml`, `cargo.lock`, and `.csproj`. For example, a tree with several `package.json` files and one `main.py` passes this check because of the Python file, but Joern selects the JavaScript frontend; with no JavaScript source, it can build an empty or irrelevant CPG and report a false-clean zero-finding scan. `.csproj` also remains in `AUTO_DETECT_PATTERNS`, so a directory containing only that manifest passes directly.

Determine the selected frontend using source-only counts and pass it explicitly via `--language`, or mirror Joern's selection first and then require at least one real source file for that selected frontend. Remove `.csproj` from the source-presence set.

dceoy (commented):
Found one remaining false-clean path in the automatic-language preflight.

.github/workflows/joern-scan.yml
dceoy: Fixed in 520642e: auto-detect now counts real-source matches per frontend itself (excluding manifests) and passes the winner explicitly via --language, and .csproj is dropped from the csharp/csharpsrc source-presence patterns.

PR #805, opened 2026-07-17
Show Trivy scan results in job summaries

dceoy:
## Summary

- preserve the existing Trivy SBOM submission to GitHub Dependency Graph
- run separate table-format Trivy reports for Docker filesystem, Docker image, and Terraform IaC scans
- publish the table output to the GitHub Actions job summary
- defer the configured Trivy failure until after the summary is written
- reuse the existing Trivy installation for the readable report

## Motivation

The current workflows use Trivy's `github` output format, which is intended for Dependency Graph submission rather than human-readable GitHub Actions output. As a result, scan findings are difficult to inspect from a workflow run.

This change separates those concerns: the existing Dependency Graph submission remains non-blocking, while table-format scans provide readable results and retain the caller-configured failure behavior.

## Impact

Workflow users will see Trivy reports directly in the job summary, including the scanned filesystem target, image reference, or Terraform search path. Existing inputs and default failure semantics remain unchanged.

Each affected workflow performs an additional Trivy scan, but the readable report skips Trivy setup and reuses the cached databases.

## Validation

- validated the modified Docker filesystem, Docker image, and Terraform workflow structures
- verified the branch only changes the three Trivy-enabled reusable workflows
- normalized Trivy action version comments to `# v0.36.0` for zizmor compatibility
- end-to-end reusable workflow execution is left to GitHub Actions CI

### DIFF
--- a/.github/workflows/docker-build-and-push.yml
+++ b/.github/workflows/docker-build-and-push.yml
@@ -123,6 +123,11 @@ on:
         type: string
         description: Timeout for the Trivy scan
         default: 5m0s
+      trivy-report-artifact-name:
+        required: false
+        type: string
+        description: Unique base name for Trivy report artifacts; omit to skip report uploads
+        default: null
       docker-metadata-action-tags:
         required: false
         type: string
@@ -214,7 +219,7 @@ jobs:
       trivy-exit-code: ${{ inputs.trivy-filesystem-exit-code }}
       trivy-config: ${{ inputs.trivy-config }}
       trivy-timeout: ${{ inputs.trivy-timeout }}
-      trivy-report-artifact-name: ${{ inputs.image-artifact-name }}-trivy-filesystem-report
+      trivy-report-artifact-name: ${{ inputs.trivy-report-artifact-name && format('{0}-filesystem', inputs.trivy-report-artifact-name) || '' }}
       runs-on: ${{ inputs.runs-on }}
     secrets:
       GH_TOKEN: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
@@ -359,7 +364,7 @@ jobs:
       trivy-exit-code: ${{ inputs.trivy-image-exit-code }}
       trivy-config: ${{ inputs.trivy-config }}
       trivy-timeout: ${{ inputs.trivy-timeout }}
-      trivy-report-artifact-name: ${{ inputs.image-artifact-name }}-trivy-image-report
+      trivy-report-artifact-name: ${{ inputs.trivy-report-artifact-name && format('{0}-image', inputs.trivy-report-artifact-name) || '' }}
       runs-on: ${{ inputs.runs-on }}
     secrets:
       GH_TOKEN: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
--- a/.github/workflows/docker-build-and-push.yml
+++ b/.github/workflows/docker-build-and-push.yml
@@ -214,6 +214,7 @@ jobs:
       trivy-exit-code: ${{ inputs.trivy-filesystem-exit-code }}
       trivy-config: ${{ inputs.trivy-config }}
       trivy-timeout: ${{ inputs.trivy-timeout }}
+      trivy-report-artifact-name: ${{ inputs.image-artifact-name }}-trivy-filesystem-report
       runs-on: ${{ inputs.runs-on }}
     secrets:
       GH_TOKEN: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
@@ -358,6 +359,7 @@ jobs:
       trivy-exit-code: ${{ inputs.trivy-image-exit-code }}
       trivy-config: ${{ inputs.trivy-config }}
       trivy-timeout: ${{ inputs.trivy-timeout }}
+      trivy-report-artifact-name: ${{ inputs.image-artifact-name }}-trivy-image-report
       runs-on: ${{ inputs.runs-on }}
     secrets:
       GH_TOKEN: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}
--- a/.github/workflows/docker-image-scan.yml
+++ b/.github/workflows/docker-image-scan.yml
@@ -103,7 +103,27 @@ jobs:
           registry: ${{ inputs.registry }}
           username: ${{ inputs.registry-user }}
           password: ${{ secrets.DOCKER_TOKEN }}  # zizmor: ignore[secrets-outside-env] caller-provided secret
-      - name: Run Trivy vulnerability scanner
+      - name: Run Trivy in GitHub SBOM mode and submit results to Dependency Graph
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25  # 0.36.0
+        with:
+          scan-type: image
+          image-ref: ${{ matrix.image-ref }}
+          scanners: ${{ inputs.trivy-scanners }}
+          severity: ${{ inputs.trivy-severity }}
+          ignore-unfixed: ${{ inputs.trivy-ignore-unfixed }}
+          exit-code: 0
+          trivy-config: ${{ inputs.trivy-config }}
+          timeout: ${{ inputs.trivy-timeout }}
+          format: github
+          output: dependency-results.${{ strategy.job-index }}.sbom.json
+          github-pat: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}  # zizmor: ignore[secrets-outside-env] caller-provided secret
+        env:
+          TRIVY_CHECKS_REPOSITORY: public.ecr.aws/aquasecurity/trivy-checks
+          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db
+          TRIVY_JAVA_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-java-db
+      - name: Run Trivy and generate a readable report
+        id: trivy-report
+        continue-on-error: true
         uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25  # 0.36.0
         with:
           scan-type: image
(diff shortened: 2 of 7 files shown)


.github/workflows/docker-image-scan.yml
```diff
@@ -121,3 +121,49 @@
           TRIVY_CHECKS_REPOSITORY: public.ecr.aws/aquasecurity/trivy-checks
           TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db
           TRIVY_JAVA_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-java-db
+      - name: Run Trivy and generate a readable report
+        id: trivy-report
+        continue-on-error: true
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25  # 0.36.0
```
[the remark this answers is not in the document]
dceoy: Fixed by changing the Trivy action version comment to `# v0.36.0`, matching the pinned tag format expected by zizmor. The existing Trivy invocation in this workflow was normalized as well.

.github/workflows/docker-lint-and-scan.yml
```diff
@@ -149,6 +149,52 @@
           TRIVY_CHECKS_REPOSITORY: public.ecr.aws/aquasecurity/trivy-checks
           TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db
           TRIVY_JAVA_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-java-db
+      - name: Run Trivy and generate a readable report
+        id: trivy-report
+        continue-on-error: true
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25  # 0.36.0
```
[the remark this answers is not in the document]
dceoy: Fixed by changing the Trivy action version comment to `# v0.36.0`, matching the pinned tag format expected by zizmor. The existing Trivy invocation in this workflow was normalized as well.

.github/workflows/docker-lint-and-scan.yml
```diff
@@ -149,6 +149,52 @@ jobs:
...
+          scan-ref: ${{ inputs.trivy-scan-ref }}
+          scanners: ${{ inputs.trivy-scanners }}
+          severity: ${{ inputs.trivy-severity }}
+          ignore-unfixed: ${{ inputs.trivy-ignore-unfixed }}
+          exit-code: ${{ inputs.trivy-exit-code }}
+          trivy-config: ${{ inputs.trivy-config }}
+          timeout: ${{ inputs.trivy-timeout }}
+          format: table
+          output: trivy-results.txt
+          skip-setup-trivy: true
```
dceoy, line 172: **[P1] Force the reporting scan to use `table` format.** `table` is the action's default, and trivy-action deliberately does not export `TRIVY_FORMAT` when an input equals its default. Therefore, when a caller supplies `trivy-config` containing another `format` (for example `github`, `json`, or `sarif`), that config value wins and `trivy-results.txt` is not a readable table. The first scan did not have this issue because `github` is non-default and is explicitly exported. Set `TRIVY_FORMAT: table` in the reporting action's `env` (or invoke the CLI with an explicit format) in all three workflows.

.github/workflows/docker-lint-and-scan.yml
```diff
@@ -149,6 +149,52 @@ jobs:
...
+            echo "**Target:** \`${TRIVY_SCAN_REF}\`"
+            echo
+            echo "<details open>"
+            echo "<summary>Scan results</summary>"
+            echo
+            echo '```text'
+            if [[ -s trivy-results.txt ]]; then
+              cat trivy-results.txt
+            else
+              echo "Trivy did not generate a report."
```
dceoy, line 211: **[P1] Bound the job-summary payload and preserve the full report separately.** GitHub limits each step summary to 1 MiB. With the current defaults (all severities plus vuln/secret/misconfig scanners), a filesystem or image report can exceed that limit. GitHub then rejects the summary upload while the step/job status remains unaffected; because Trivy writes findings only to `trivy-results.txt`, users can get a failed scan with no readable findings in either the summary or logs. Truncate the summary below 1 MiB and upload the complete report as an artifact (or otherwise provide a durable link). Apply the same handling to all three workflows.

dceoy (commented):
I found two correctness gaps that can prevent the new job-summary reporting from working as described. Both patterns are duplicated across the three modified reusable workflows, so please apply the fixes consistently to Docker filesystem, Docker image, and Terraform scans.

dceoy (commented):
# PR Review Summary

No additional actionable issues meeting the review’s reporting threshold were identified after deduplicating against the existing review discussion.

## Checked

- The complete three-workflow diff for Docker filesystem, Docker image, and Terraform Trivy scans.
- Preservation of Dependency Graph submission, deferred failure behavior, report generation, summary rendering, caller-provided configuration, action pinning, and security boundaries around workflow inputs.
- Existing review submissions and all review threads.
- Current-head Snyk status succeeded for `a657fe3e757f920707530c436ed34d4b358885bb`.

## Notes

- Two unresolved threads already cover the material reporting gaps: caller `trivy-config` can override the intended table format, and an unbounded report can exceed GitHub’s step-summary size limit. They are not repeated because no new evidence changes those assessments.
- End-to-end reusable-workflow execution was not available as a current-head status through the connector.

## Recommended Action

Address the two existing unresolved reporting findings consistently across all three workflows before merge. No separate new finding is added by this review.

.github/workflows/docker-lint-and-scan.yml
```diff
@@ -149,6 +149,64 @@ jobs:
...
+          TRIVY_CHECKS_REPOSITORY: public.ecr.aws/aquasecurity/trivy-checks
+          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db
+          TRIVY_JAVA_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-java-db
+      - name: Upload complete Trivy report
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
+        with:
+          name: trivy-filesystem-report
+          path: trivy-results.txt
+          if-no-files-found: ignore
```
dceoy, line 184: **[P1] Make report artifact names unique across reusable-workflow calls.** `upload-artifact@v4+` treats artifact names as immutable and unique within the caller's workflow run. A caller can legally invoke this reusable workflow more than once (including through a job matrix), but every invocation currently uploads `trivy-filesystem-report`; the second upload will fail with an artifact conflict and make an otherwise successful scan job fail. The same issue applies to `trivy-terraform-report`, and `trivy-image-report-${{ strategy.job-index }}` still collides across separate invocations because each invocation starts its index at 0. Add a caller-provided artifact-name/suffix input or otherwise include a per-invocation unique component in all three artifact names; avoid `overwrite: true`, which would race and discard another report.

dceoy (commented):
The current reporting and deferred-failure logic is otherwise consistent, and the previously raised format/size issues are fixed. I found one remaining caller-level compatibility issue with the new artifacts that should be addressed across all three workflows.

.github/workflows/docker-lint-and-scan.yml
```diff
@@ -53,6 +53,10 @@ on:
         type: string
         description: Timeout for the Trivy scan
         default: 5m0s
+      trivy-report-artifact-name:
+        required: true
+        type: string
+        description: Unique name for the Trivy report artifact
```
dceoy: **[P1] Preserve the existing `workflow_call` contract.** This is a newly required input on a publicly advertised reusable workflow, so an existing direct caller that previously relied only on defaults will fail workflow validation before any job starts. The same breaking change exists in `docker-image-scan.yml`, `terraform-lint-and-scan.yml`, and `terraform-lock-files-upgrade-and-pr-merge.yml`; for the Terraform workflows it is required even when Trivy is disabled. Make the input optional with a backward-compatible default/conditional upload, or explicitly ship this as a major-version API break with migration documentation.

.github/workflows/docker-build-and-push.yml
```diff
@@ -214,6 +214,7 @@ jobs:
       trivy-exit-code: ${{ inputs.trivy-filesystem-exit-code }}
       trivy-config: ${{ inputs.trivy-config }}
       trivy-timeout: ${{ inputs.trivy-timeout }}
+      trivy-report-artifact-name: ${{ inputs.image-artifact-name }}-trivy-filesystem-report
```
dceoy: **[P1] The wrapper still does not make report artifact names unique.** `image-artifact-name` defaults to `docker-image`, so two caller jobs can invoke this workflow with defaults and both attempt to upload `docker-image-trivy-filesystem-report` (and later the same image-report name). `upload-artifact@v4+` rejects the second upload, turning a previously valid multi-call workflow into a failure. Expose a dedicated caller-level report name/suffix that must differ per invocation, or provide another collision-free scheme; apply the same fix to the buildx-bake and AWS-pull wrappers.

dceoy (commented):
The readable-report and deferred-failure implementation is otherwise coherent, and the earlier format and summary-size findings are addressed. Two compatibility issues remain and should be fixed before merge: the new required input breaks existing direct callers, and the wrapper-derived artifact names still collide when the wrapper is invoked multiple times with its defaults.

.github/workflows/docker-lint-and-scan.yml
```diff
@@ -149,6 +154,64 @@ jobs:
...
+            echo "<summary>Scan results</summary>"
+            echo
+            echo '```text'
+            if [[ -s trivy-results.txt ]]; then
+              head -c 900000 trivy-results.txt
+              if [[ $(wc -c < trivy-results.txt) -gt 900000 ]]; then
+                echo
+                echo "[Report truncated; download the complete report artifact.]"
+              fi
+            else
```
dceoy, line 210: **[P2] Do not direct users to an artifact that is disabled by default.** When `trivy-results.txt` exceeds 900,000 bytes and `trivy-report-artifact-name` is omitted (the backward-compatible default), this message says to download the complete artifact even though no artifact was uploaded, and the omitted tail is unavailable from either the summary or Trivy logs. Either emit the complete report to the log in this case, or vary the notice to explain that callers must set `trivy-report-artifact-name`; apply the same handling to the image and Terraform summaries.

dceoy (commented):
The earlier format, summary-size, artifact-collision, and backward-compatibility issues are addressed consistently, and the current CI passes. I found one remaining report-accessibility gap in the default (artifact upload disabled) path.

dceoy (commented):
# PR Review Summary

No actionable issues meeting the review’s reporting threshold were identified.

## Checked

- The current seven-file workflow diff and the new caller-facing `trivy-report-artifact-name` propagation.
- Trivy Dependency Graph submission, explicit table-format reporting, deferred failure enforcement, bounded job-summary output, optional full-report artifacts, and matrix artifact naming.
- Backward compatibility of reusable-workflow inputs and default artifact-disabled behavior.
- All existing review submissions and review threads; the earlier format, summary-size, artifact-collision, required-input, and truncation-message findings are resolved in the current diff.
- Current-head Snyk status for `6724f71bfb18102188597a8c7e462e0e0cdff34c`.

## Notes

End-to-end reusable-workflow execution was not exposed as a current-head check through the connector; only the successful Snyk status was available.

## Recommended Action

No automated blocking concerns were identified. Merge remains a maintainer decision after required checks pass.

PR #826, opened 2026-07-25
Scope AWS credentials to dependent steps

### LINKED ISSUE #807: Support step-scoped AWS credentials
## Summary

Support step-scoped AWS credentials by using the outputs of `aws-actions/configure-aws-credentials` instead of exporting temporary AWS credentials as job-wide environment variables.

## Background

Most AWS workflows currently use `aws-actions/configure-aws-credentials`, which exports temporary AWS credentials to the environment for all subsequent steps by default.

Many workflows only require AWS credentials in a small number of steps. Limiting credential exposure to the steps that actually require them reduces the credential exposure surface and makes credential usage explicit.

`aws-actions/configure-aws-credentials` supports this pattern through:

- `output-credentials: true`
- `output-env-credentials: false`

The resulting credentials can then be passed explicitly to AWS-dependent steps.

## Proposed changes

- Use step-scoped AWS credentials where practical.
- Configure `aws-actions/configure-aws-credentials` with `output-credentials: true` and `output-env-credentials: false`.
- Pass the following outputs only to steps that require AWS access:
  - `AWS_ACCESS_KEY_ID`
  - `AWS_SECRET_ACCESS_KEY`
  - `AWS_SESSION_TOKEN`
  - `AWS_REGION`
- Avoid applying the pattern where it would introduce substantial YAML duplication without meaningfully reducing exposure.

## Candidate workflows

Good initial candidates:

- `.github/workflows/aws-parameter-store-update.yml`
- `.github/workflows/aws-codebuild-run.yml`

Requires further evaluation:

- `.github/workflows/docker-pull-from-aws.yml`

Likely lower priority:

- Terraform and Terragrunt workflows, where most subsequent steps require AWS credentials and explicit propagation would significantly increase duplication.

## Benefits

- Reduce unnecessary exposure of temporary AWS credentials.
- Make AWS credential use explicit at the step level.
- Improve least-privilege practices within GitHub Actions jobs.
- Reduce the risk of unrelated commands or third-party actions inheriting AWS credentials.

## Reference

- https://github.com/aws-actions/configure-aws-credentials

dceoy:
## Summary

- configure every `aws-actions/configure-aws-credentials` step to emit credential outputs without exporting job-wide AWS environment variables
- pass credentials and the resolved AWS region only to AWS-dependent Parameter Store, CodeBuild, ECR, Claude, and PR-agent steps
- scope credentials to each Terraform and Terragrunt init, validation, plan, apply, and destroy step that may access AWS

## Validation

- `scripts/qa.sh`

Closes #807

### DIFF
--- a/.github/workflows/aws-codebuild-run.yml
+++ b/.github/workflows/aws-codebuild-run.yml
@@ -144,14 +144,22 @@ jobs:
         run: |
           cat "${AWS_PROFILE_ENV_FILE}" >> "${GITHUB_OUTPUT}"
       - name: Configure AWS credentials
+        id: aws-credentials
         uses: aws-actions/configure-aws-credentials@517a711dbcd0e402f90c77e7e2f81e849156e31d  # v6.2.2
         with:
           role-to-assume: ${{ inputs.aws-iam-role-to-assume || steps.aws-profile-env.outputs.ROLE_ARN || null }}
           aws-region: ${{ inputs.aws-region || steps.aws-profile-env.outputs.REGION || null }}
           role-session-name: github-actions-${{ github.run_id }}
+          output-credentials: true
+          output-env-credentials: false
       - name: Run an AWS CodeBuild project
         id: aws-codebuild-run-build
         uses: aws-actions/aws-codebuild-run-build@7e46c3fa1c1f217e26a73712796b1f78938b534b  # v1.0.19
+        env:
+          AWS_ACCESS_KEY_ID: ${{ steps.aws-credentials.outputs.aws-access-key-id }}
+          AWS_SECRET_ACCESS_KEY: ${{ steps.aws-credentials.outputs.aws-secret-access-key }}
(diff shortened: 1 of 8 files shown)


dceoy (commented):
No blocking issues found.

The implementation follows the documented `aws-actions/configure-aws-credentials` pattern for step-scoped credentials: `output-credentials: true` exposes masked step outputs, while `output-env-credentials: false` prevents exporting the credentials job-wide. The downstream AWS-dependent steps explicitly receive the access key, secret key, session token, and region.

The existing OpenCode comment appears to be a false positive: with `output-env-credentials: false`, the action does not write these credentials to `GITHUB_ENV`; changing `output-credentials` to `false` would also make the referenced step outputs unavailable.

CI linting and security scans are passing. The remaining limitation is that these checks do not exercise the reusable workflows against AWS, but I do not consider that a merge blocker for this change.

PR #827, opened 2026-07-25
Fix reusable workflow regressions and agent docs

### LINKED ISSUE #736: docker-pull-from-aws: image_refs_json is never written to GITHUB_OUTPUT, so postpull-scan fails on every run (regression from #652 fix)
## Problem

In `.github/workflows/docker-pull-from-aws.yml`, the `Output the image references` step prints the `image_refs_json=...` line to stdout but never appends it to `${GITHUB_OUTPUT}`. The job output `image-refs-json` is therefore always empty, and the `postpull-scan` job — enabled by default via `scan-after-pull: true` — fails at job setup because its matrix evaluates `fromJSON('')`.

This appears to be a regression introduced while reworking this workflow's shell pipelines for #652: every other output-producing step in the file uses `| tee -a "${GITHUB_OUTPUT}"`, but this one lost it.

## Evidence

`.github/workflows/docker-pull-from-aws.yml` (at cd07701):

```yaml
      - name: Output the image references
        id: output-image-refs
        run: |
          jq -rRs 'split("\n")[:-1] | "image_refs_json=\(.)"' < "${IMAGE_URI_LIST_TXT}"   # lines 150-153: stdout only
```

- Lines 111-112: `outputs: image-refs-json: ${{ steps.output-image-refs.outputs.image_refs_json }}` → always empty.
- Lines 173-180: `postpull-scan` passes `image-refs-json: ${{ needs.pull-and-save.outputs.image-refs-json }}` (empty string) to `docker-image-scan.yml`.
- `.github/workflows/docker-image-scan.yml:79`: `matrix: image-ref: ${{ fromJSON(inputs.image-refs-json) }}` — `fromJSON('')` is an evaluation error, so the `scan` job fails before any step runs.

Compare the same pattern done correctly elsewhere, e.g. `github-pr-branch-aggregation.yml:104-111` or `docker-buildx-bake.yml:268-273`, which `tee -a "${GITHUB_OUTPUT}"`.

## Impact

- Any caller of `docker-pull-from-aws.yml` with default inputs gets a failed workflow run (the pull/save part succeeds, then `postpull-scan` fails).
- `docker-save-and-terraform-deploy-to-aws.yml` (lines 286-310) invokes this workflow as its `pull` job with `scan-after-pull` defaulting to `true`; the resulting failure also skips the downstream `deploy` job (`if: (! (failure() || cancelled()))`), breaking the pull→deploy path of that composed pipeline.
- Related: #672 asks for `image-refs-json` to be exposed as a `workflow_call` output — that will also silently return an empty value until this is fixed.

## Suggested fix

Append the output to `GITHUB_OUTPUT`:

```yaml
      - name: Output the image references
        id: output-image-refs
        run: |
          jq -rRs 'split("\n")[:-1] | "image_refs_json=\(.)"' < "${IMAGE_URI_LIST_TXT}" \
            | tee -a "${GITHUB_OUTPUT}"
```

Also consider moving this step after `Tag the Docker images` (lines 154-162), which appends the retagged target references to `${IMAGE_URI_LIST_TXT}` — as written, even with the fix, target tags are saved into the tarball but excluded from the scan matrix. Deciding whether that is intended should be part of the fix.

## Validation

- Call `docker-pull-from-aws.yml` with an `image-names` value and default `scan-after-pull`; the `postpull-scan` matrix should materialize with the pulled refs and the run should complete.
- Check the run's step summary: the `output-image-refs` step must show the `image_refs_json=[...]` line being written to `GITHUB_OUTPUT` (visible via `tee`).

### LINKED ISSUE #737: docker-build-and-push: cosign verify matches the caller repository, but Fulcio certificates carry this repo's job_workflow_ref — verification fails for external callers
## Problem

The `Verify the image signatures` step added for #654 builds its `--certificate-identity-regexp` from `${GITHUB_REPOSITORY}`, i.e. the **calling** repository. For keyless signing performed inside a job that is defined in a reusable workflow, the Fulcio certificate's SAN is derived from the OIDC token's `job_workflow_ref` claim — which points at the repository that **defines** the reusable workflow (`dceoy/gh-actions-for-devops`), not the caller. So whenever another repository calls this workflow with `cosign: true` and `push: true`, signing succeeds but the verification step fails, failing the whole job.

It only appears to work when this repo calls its own workflow, because then the two identities coincide.

## Evidence

`.github/workflows/docker-build-and-push.yml:291-303` (at cd07701):

```yaml
      - name: Verify the image signatures
        if: inputs.cosign && inputs.push
        ...
        run: |
          while IFS= read -r tag; do
            [[ -n "${tag}" ]] || continue
            cosign verify \
              --certificate-identity-regexp "https://github.com/${GITHUB_REPOSITORY}/.github/workflows/.+@refs/.+" \
```

For a caller `acme/app` using `dceoy/gh-actions-for-devops/.github/workflows/docker-build-and-push.yml@v…`, the certificate SAN is
`https://github.com/dceoy/gh-actions-for-devops/.github/workflows/docker-build-and-push.yml@refs/…`, while the regexp requires `https://github.com/acme/app/...` — no match, `cosign verify` exits non-zero.

References: [sigstore/cosign discussion #2936](https://github.com/sigstore/cosign/discussions/2936) (certificate identity for reusable workflows is the reusable workflow ref); the same convention is what slsa-framework/slsa-github-generator relies on for verification.

Affected pass-throughs:
- `docker-build-with-multi-targets.yml:235` (`cosign: ${{ inputs.cosign }}`)
- `docker-save-and-terraform-deploy-to-aws.yml:264` (`cosign: ${{ inputs.cosign }}`)

## Impact

The `cosign: true` feature of this catalog's flagship build workflow is unusable from any consuming repository: the job fails after images were already pushed and signed. (Fails closed, so no integrity risk — but the feature is broken for exactly its intended audience.)

## Suggested fix

Match the identity that Fulcio actually certifies — this reusable workflow:

```yaml
          cosign verify \
            --certificate-identity-regexp "https://github.com/dceoy/gh-actions-for-devops/\.github/workflows/docker-build-and-push\.yml@refs/.+" \
            --certificate-oidc-issuer https://token.actions.githubusercontent.com \
```

If the repo should stay fork-friendly, derive the value instead of hardcoding: expose an optional `cosign-certificate-identity-regexp` input, and/or extract the identity from the OIDC claims at runtime (e.g. decode `job_workflow_ref` from a requested token) rather than `GITHUB_REPOSITORY`. Documenting the certified identity also helps consumers write their own `cosign verify`/policy-controller rules.

## Validation

- From a separate repository, call `docker-build-and-push.yml` with `cosign: true`, `push: true` against a scratch image; the verify step must pass.
- Inspect the certificate on a signed image (`cosign verify … | jq`, or `crane`/`rekor-cli`) and confirm the SAN is the reusable-workflow ref, matching the new regexp.

### LINKED ISSUE #738: docker-build-and-push: default `registry: null` yields `images: /owner/name` — invalid Docker references break the no-registry (push:false) path
## Problem

`docker-build-and-push.yml` unconditionally concatenates `images: ${{ inputs.registry }}/${{ inputs.image-name }}`. With the default `registry: null` (rendered as an empty string), `docker/metadata-action` receives an image name with a leading slash (e.g. `/acme/app`). metadata-action performs no normalization or validation beyond non-empty (checked in its `src/image.ts`), so it emits tags like `/acme/app:latest`, which are invalid Docker references — the subsequent `docker/build-push-action` tagging fails with `invalid reference format`.

So the workflow cannot run without a registry, even though `registry` is optional and the tarball-artifact flow (`push: false`, image travels as an artifact) is explicitly designed for that case.

## Evidence

`.github/workflows/docker-build-and-push.yml` (at cd07701):

- Lines 6-10: `registry` input, `default: null`.
- Line 255: `images: ${{ inputs.registry }}/${{ inputs.image-name }}` → `/owner/name` when registry is unset.
- Lines 264-283: `build-push-action` consumes `steps.meta.outputs.tags` directly.

Reachable from this repo's own composition: `docker-save-and-terraform-deploy-to-aws.yml:252-262` calls `docker-build-with-multi-targets.yml` with `push: false` and `registry: ${{ inputs.docker-registry }}` (default `null`), and `docker-build-with-multi-targets.yml:223-225` forwards `registry`/`image-name` per matrix target into this workflow — so the local-build → tarball → Terraform-deploy pipeline hits the invalid reference whenever `docker-registry` is omitted.

Related but distinct from #669 (item 2: multi-arch `--load` also breaks the `push: false` default path) and closed #653 (empty `image-name` from matrix keys).

## Impact

Any invocation without `registry` fails during build/tag. The failure message (`invalid reference format`) doesn't point at the actual cause (input concatenation), costing consumers debugging time. Defaults as shipped are self-contradictory: `push: true` + `registry: null` can never succeed.

## Suggested fix

Only prefix the registry when it is set:

```yaml
      - name: Extract metadata from Git reference and GitHub events
        id: meta
        uses: docker/metadata-action@...
        with:
          images: ${{ inputs.registry && format('{0}/{1}', inputs.registry, inputs.image-name) || inputs.image-name }}
```

Apply the same guard anywhere the pattern recurs in the catalog. Optionally fail fast with a clear error when `push: true` and `registry` is empty.

## Validation

- Call the workflow with `push: false`, no `registry`, single platform: metadata tags should be `owner/name:…` and the build/save/upload steps should succeed.
- Call with `registry: ghcr.io`: tags remain `ghcr.io/owner/name:…` (no behavior change).

### LINKED ISSUE #739: python-package-release-on-pypi-and-github: version-tag check fails for any project name that normalizes in dist filenames (hyphens, dots, uppercase)
## Problem

The `Validate the version-tag consistency` step looks for dist files with `find dist -name "${PROJECT_NAME}-*"`, where `PROJECT_NAME` is the raw `project.name` read from `pyproject.toml`. Per PEP 427/PEP 625, build backends (setuptools, hatchling, poetry-core, uv) normalize the distribution name in wheel **and** sdist filenames: lowercase, with runs of `-`, `_`, `.` collapsed to `_`. So for any project named e.g. `my-package`, the files are `my_package-1.2.3.tar.gz` / `my_package-1.2.3-py3-none-any.whl`, the `find` matches nothing, `v` ends up empty, and the check always fails — blocking the release even when the tag is correct.

Additionally, `cut -d '-' -f 2` assumes the filename's first `-` terminates the name, so even a legacy backend that kept a hyphenated name in filenames (`my-package-1.2.3.tar.gz`) would extract `package` instead of the version.

## Evidence

`.github/workflows/python-package-release-on-pypi-and-github.yml:225-238` (at cd07701):

```yaml
      - name: Validate the version-tag consistency
        env:
          TAG_NAME: ${{ github.ref_name }}
          NEEDS_BUILD_OUTPUTS_PROJECT_NAME: ${{ needs.build.outputs.project-name }}
        run: |
          v="$( \
            find dist -type f -name "${NEEDS_BUILD_OUTPUTS_PROJECT_NAME}-*" -exec basename {} \; \
              | head -n 1 \
              | cut -d '-' -f 2 \
          )"
          v="${v%.tar.gz}"
          if [[ "${TAG_NAME}" != "${v}" ]] && [[ "${TAG_NAME}" != "v${v}" ]]; then
            echo "The tag (${TAG_NAME}) is inconsistent with the version (${v})." && exit 1
```

The name comes unmodified from `pyproject.toml` (`read-project-name` step, lines 136-153). Reproduce locally:

```console
$ mkdir -p dist && touch dist/my_package-1.2.3.tar.gz
$ NAME=my-package; find dist -type f -name "${NAME}-*"
$ # no output → v="" → check fails for tag v1.2.3
```

## Impact

Releases fail with a misleading "tag is inconsistent with the version ()" error for the large class of PyPI projects whose names contain `-`, `.`, or uppercase (hyphenated names are extremely common). The workflow only works for names that are already in normalized filename form.

## Suggested fix

Normalize the name the same way build backends do, and parse the version by stripping the name prefix instead of counting `-` fields:

```bash
name_glob="$(tr '[:upper:]' '[:lower:]' <<< "${NEEDS_BUILD_OUTPUTS_PROJECT_NAME}" | sed -E 's/[-_.]+/_/g')"
first_dist="$(find dist -type f -name "${name_glob}-*" -exec basename {} \; | head -n 1)"
v="${first_dist#"${name_glob}"-}"   # then strip .tar.gz / -py3-none-any.whl suffixes
v="${v%%-*}"
v="${v%.tar.gz}"
```

(Equivalently, read the version from `pyproject.toml`/`importlib.metadata` in the existing Python step and compare that against the tag, avoiding filename parsing entirely.)

## Validation

- Fixture package named `my-package`, tag `v1.2.3`: the check must pass after building with uv/poetry/build.
- Negative case: tag `v1.2.4` against version `1.2.3` must still fail.
- Names with dots/uppercase (e.g. `My.Package`) behave the same.

### LINKED ISSUE #740: python-package-release-on-pypi-and-github: TestPyPI job relies on trusted publishing, which PyPI rejects inside reusable workflows — `publish-to-testpypi-before-pypi: true` can never succeed and blocks the PyPI publish
## Problem

The `publish-to-testpypi` job invokes `pypa/gh-action-pypi-publish` without a `password`, i.e. via OIDC **trusted publishing**. But PyPI/TestPyPI reject trusted publishing when the publish step runs inside a reusable workflow (`invalid-publisher`: the token's `job_workflow_ref` points at this repo's workflow, which cannot match a publisher configured on the consumer's project). The same file already acknowledges this constraint for the PyPI job, which uses an API token instead — with the inline note "Reusable workflow supports only API-token-based PyPI publishing."

So for any external caller, enabling `publish-to-testpypi-before-pypi: true` produces a job that always fails — and because `publish-to-pypi` has `needs: [build, publish-to-testpypi, github-release]` with `if: … && (! (failure() || cancelled()))`, the TestPyPI failure also skips the real PyPI publish.

## Evidence

`.github/workflows/python-package-release-on-pypi-and-github.yml` (at cd07701):

- Lines 183-206: `publish-to-testpypi` — `pypa/gh-action-pypi-publish` with only `repository-url: https://test.pypi.org/legacy/`; no `password`, `permissions: id-token: write` → trusted-publishing path.
- Lines 282-286: `publish-to-pypi` — `password: ${{ secrets.PYPI_API_TOKEN }}   # zizmor: ignore[use-trusted-publishing] Reusable workflow supports only API-token-based PyPI publishing.`
- Lines 260-269: `publish-to-pypi` `needs`/`if` chain that turns a TestPyPI failure into a skipped PyPI publish.

Upstream references: [pypa/gh-action-pypi-publish#166 — "Reusable workflows are currently unsupported on the PyPI side (invalid-publisher)"](https://github.com/pypa/gh-action-pypi-publish/issues/166), [pypi/warehouse#11096](https://github.com/pypi/warehouse/issues/11096).

## Impact

- A documented input combination (`publish-to-pypi: true`, `publish-to-testpypi-before-pypi: true`) is broken by design for the workflow's intended audience (external callers), and its failure mode also cancels the production PyPI upload.
- The two publish jobs are inconsistent: one assumes token auth is the only option, the other assumes trusted publishing works.

## Suggested fix

Make TestPyPI use the same token mechanism as PyPI:

- Add an optional secret (e.g. `TEST_PYPI_API_TOKEN`) and pass it as `password:` in the TestPyPI job; fail fast with a clear message when `publish-to-testpypi-before-pypi: true` but the secret is absent.
- Alternatively, drop the TestPyPI option (YAGNI) or document that it only works when this workflow is vendored into the calling repository.

(If trusted publishing is the long-term goal, the upstream-recommended layout is to keep build/sign in the reusable workflow and do the publish step in a job of the caller's top-level workflow — worth documenting for consumers.)

## Validation

- From a consumer repo with `publish-to-testpypi-before-pypi: true` and a TestPyPI token configured, a tag release publishes to TestPyPI, then PyPI.
- Without the TestPyPI secret, the run fails early with an actionable error rather than `invalid-publisher`.

### LINKED ISSUE #742: AGENTS.md still documents the removed `.opencode/` directory (deleted in #723), misdirecting agents and contributors
## Problem

AGENTS.md's "Project Structure & Module Organization" section states:

> OpenCode review subagents and the `/review-pr` command live under `.opencode/` (auto-discovered by OpenCode), converted from Anthropic's `pr-review-toolkit` plugin.

The `.opencode/` directory no longer exists — it was removed in "Remove OpenCode custom agent definitions" (#723, commit 15fb644). Since `CLAUDE.md` is a symlink to `AGENTS.md`, this stale paragraph is injected as authoritative instructions into every AI-agent session (Claude Code, Cursor, OpenCode) working on this repo, sending them to a nonexistent path.

## Evidence

- `AGENTS.md`, first paragraph of "Project Structure & Module Organization" (at cd07701).
- `git log --oneline -- .opencode` → `15fb644 Remove OpenCode custom agent definitions (#723)`; `ls .opencode` → does not exist; `.opencode` absent from `git ls-files`.
- `/review-pr` is now supplied at runtime instead: `claude-code-review.yml:85-88` uses the `pr-review-toolkit@claude-plugins-official` plugin, and `opencode-review.yml:28-29,58` gets it from the opencode-action's bundled toolkit (`enable-toolkit`).

## Impact

Agent sessions and new contributors are pointed at a directory that doesn't exist, and the described mechanism ("auto-discovered by OpenCode") no longer reflects how the review commands are provided. For a file whose whole purpose is to keep automated contributors accurate, this is a correctness bug in the docs.

## Suggested fix

Update the sentence to describe the current state, e.g.: review tooling comes from the `pr-review-toolkit` plugin (Claude) and the opencode-action bundled toolkit (OpenCode), configured in `.github/workflows/claude-code-review.yml` and `.github/workflows/opencode-review.yml`; local agent skills live under `.agents/skills/` (symlinked from `.claude/skills`).

## Validation

- Every path named in AGENTS.md exists in the repo (`git ls-files` cross-check).
- A fresh agent session reading CLAUDE.md finds no references to removed directories.

dceoy:
## Summary

- emit Docker pull image references through `GITHUB_OUTPUT` after target tags are added
- make Docker metadata work without a registry and verify cosign signatures against the reusable workflow identity
- support normalized Python distribution names and API-token authentication for TestPyPI
- replace the stale `.opencode/` guidance with the current review workflow locations

## Checks

- `.github/scripts/update-readme.sh`
- `scripts/qa.sh`
- targeted shell checks for normalized distribution names and image reference JSON

Closes #736
Closes #737
Closes #738
Closes #739
Closes #740
Closes #742

### DIFF
--- a/.github/workflows/docker-build-and-push.yml
+++ b/.github/workflows/docker-build-and-push.yml
@@ -258,7 +258,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@dc802804100637a589fabce1cb79ff13a1411302  # v6.2.0
         with:
-          images: ${{ inputs.registry }}/${{ inputs.image-name }}
+          images: ${{ inputs.registry && format('{0}/{1}', inputs.registry, inputs.image-name) || inputs.image-name }}
           tags: ${{ inputs.docker-metadata-action-tags }}
       - name: Login to the Docker registry
         if: inputs.registry != null
@@ -303,7 +303,7 @@ jobs:
           while IFS= read -r tag; do
             [[ -n "${tag}" ]] || continue
             cosign verify \
-              --certificate-identity-regexp "https://github.com/${GITHUB_REPOSITORY}/.github/workflows/.+@refs/.+" \
+              --certificate-identity-regexp "https://github.com/dceoy/gh-actions-for-devops/\\.github/workflows/docker-build-and-push\\.yml@refs/.+" \
               --certificate-oidc-issuer https://token.actions.githubusercontent.com \
               "${tag}@${DIGEST}"
           done <<< "${TAGS}"
--- a/.github/workflows/docker-build-and-push.yml
+++ b/.github/workflows/docker-build-and-push.yml
@@ -303,7 +303,7 @@ jobs:
           while IFS= read -r tag; do
             [[ -n "${tag}" ]] || continue
             cosign verify \
-              --certificate-identity-regexp "https://github.com/dceoy/gh-actions-for-devops/\\.github/workflows/docker-build-and-push\\.yml@refs/.+" \
+              --certificate-identity-regexp "https://github.com/dceoy/gh-actions-for-devops/\\.github/workflows/docker-build-and-push\\.yml@(refs/.+|[0-9a-f]{40})" \
               --certificate-oidc-issuer https://token.actions.githubusercontent.com \
               "${tag}@${DIGEST}"
           done <<< "${TAGS}"
--- a/.github/workflows/python-package-release-on-pypi-and-github.yml
+++ b/.github/workflows/python-package-release-on-pypi-and-github.yml
@@ -84,6 +84,9 @@ on:
       PYPI_API_TOKEN:
         required: false
         description: PyPI API token
+      TEST_PYPI_API_TOKEN:
+        required: false
+        description: TestPyPI API token
       GH_TOKEN:
         required: false
         description: GitHub token
@@ -193,16 +196,25 @@ jobs:
       name: testpypi
       url: https://test.pypi.org/p/${{ needs.build.outputs.project-name }}
     permissions:
-      id-token: write  # IMPORTANT: mandatory for trusted publishing
+      contents: read
     steps:
       - name: Download all the dists
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
         with:
           name: ${{ inputs.distribution-artifact-name }}
           path: dist/
+      - name: Validate TestPyPI API token
+        env:
+          TEST_PYPI_API_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        run: |
+          if [[ -z "${TEST_PYPI_API_TOKEN}" ]]; then
+            echo "TEST_PYPI_API_TOKEN is required when publish-to-testpypi-before-pypi is enabled." >&2
+            exit 1
+          fi
       - name: Publish distribution 📦 to TestPyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
         with:
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}  # zizmor: ignore[use-trusted-publishing] Reusable workflow supports only API-token-based TestPyPI publishing.
           repository-url: https://test.pypi.org/legacy/
   github-release:
     name: Sign the Python 🐍 distribution 📦 with Sigstore and upload them to GitHub Release
@@ -227,11 +239,21 @@ jobs:
           TAG_NAME: ${{ github.ref_name }}
           NEEDS_BUILD_OUTPUTS_PROJECT_NAME: ${{ needs.build.outputs.project-name }}
         run: |
-          v="$( \
-            find dist -type f -name "${NEEDS_BUILD_OUTPUTS_PROJECT_NAME}-*" -exec basename {} \; \
-              | head -n 1 \
-              | cut -d '-' -f 2 \
+          distribution_name="$( \
+            tr '[:upper:]' '[:lower:]' <<< "${NEEDS_BUILD_OUTPUTS_PROJECT_NAME}" \
+              | sed -E 's/[-_.]+/_/g' \
           )"
+          first_dist="$( \
+            find dist -type f -name "${distribution_name}-*" -exec basename {} \; \
+              | sort \
+              | head -n 1 \
+          )"
+          if [[ -z "${first_dist}" ]]; then
+            echo "No distribution file found for ${NEEDS_BUILD_OUTPUTS_PROJECT_NAME}." >&2
+            exit 1
+          fi
+          v="${first_dist#"${distribution_name}"-}"
+          v="${v%%-*}"
           v="${v%.tar.gz}"
           if [[ "${TAG_NAME}" != "${v}" ]] && [[ "${TAG_NAME}" != "v${v}" ]]; then
             echo "The tag (${TAG_NAME}) is inconsistent with the version (${v})." && exit 1
@@ -272,7 +294,7 @@ jobs:
       name: pypi
       url: https://pypi.org/p/${{ needs.build.outputs.project-name }}
     permissions:
-      id-token: write  # IMPORTANT: mandatory for trusted publishing
+      contents: read
     steps:
       - name: Download all the dists
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
--- a/.github/workflows/python-package-release-on-pypi-and-github.yml
+++ b/.github/workflows/python-package-release-on-pypi-and-github.yml
@@ -207,6 +207,16 @@ jobs:
       # trusted publishing, which upstream does not yet support for an
       # externally defined reusable workflow (pypa/gh-action-pypi-publish#166).
       # Callers must set TEST_PYPI_API_TOKEN to publish successfully.
+      - name: Check the TestPyPI API token is set
+        env:
+          TEST_PYPI_API_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        run: |
+          if [[ -z "${TEST_PYPI_API_TOKEN}" ]]; then
+            echo "TEST_PYPI_API_TOKEN is required when publish-to-testpypi-before-pypi is enabled" \
+              "because trusted publishing is not supported for this externally defined reusable" \
+              "workflow (pypa/gh-action-pypi-publish#166). Set the TEST_PYPI_API_TOKEN secret." >&2
+            exit 1
+          fi
       - name: Publish distribution 📦 to TestPyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
         with:
--- a/.github/workflows/python-package-release-on-pypi-and-github.yml
+++ b/.github/workflows/python-package-release-on-pypi-and-github.yml
@@ -203,6 +203,10 @@ jobs:
         with:
           name: ${{ inputs.distribution-artifact-name }}
           path: dist/
+      # NOTE: Without TEST_PYPI_API_TOKEN, this action falls back to OIDC
+      # trusted publishing, which upstream does not yet support for an
+      # externally defined reusable workflow (pypa/gh-action-pypi-publish#166).
+      # Callers must set TEST_PYPI_API_TOKEN to publish successfully.
       - name: Publish distribution 📦 to TestPyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
         with:
@@ -293,6 +297,10 @@ jobs:
         with:
           name: ${{ inputs.distribution-artifact-name }}
           path: dist/
+      # NOTE: Without PYPI_API_TOKEN, this action falls back to OIDC
+      # trusted publishing, which upstream does not yet support for an
+      # externally defined reusable workflow (pypa/gh-action-pypi-publish#166).
+      # Callers must set PYPI_API_TOKEN to publish successfully.
       - name: Publish distribution 📦 to PyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
         with:
--- a/.github/workflows/python-package-release-on-pypi-and-github.yml
+++ b/.github/workflows/python-package-release-on-pypi-and-github.yml
@@ -196,25 +196,17 @@ jobs:
       name: testpypi
       url: https://test.pypi.org/p/${{ needs.build.outputs.project-name }}
     permissions:
-      contents: read
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
     steps:
       - name: Download all the dists
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
         with:
           name: ${{ inputs.distribution-artifact-name }}
           path: dist/
-      - name: Validate TestPyPI API token
-        env:
-          TEST_PYPI_API_TOKEN: ${{ secrets.TEST_PYPI_API_TOKEN }}
-        run: |
-          if [[ -z "${TEST_PYPI_API_TOKEN}" ]]; then
-            echo "TEST_PYPI_API_TOKEN is required when publish-to-testpypi-before-pypi is enabled." >&2
-            exit 1
-          fi
       - name: Publish distribution 📦 to TestPyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
         with:
-          password: ${{ secrets.TEST_PYPI_API_TOKEN }}  # zizmor: ignore[use-trusted-publishing] Reusable workflow supports only API-token-based TestPyPI publishing.
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}  # zizmor: ignore[use-trusted-publishing] Optional API-token authentication.
           repository-url: https://test.pypi.org/legacy/
   github-release:
     name: Sign the Python 🐍 distribution 📦 with Sigstore and upload them to GitHub Release
@@ -294,7 +286,7 @@ jobs:
       name: pypi
       url: https://pypi.org/p/${{ needs.build.outputs.project-name }}
     permissions:
-      contents: read
+      id-token: write  # IMPORTANT: mandatory for trusted publishing
     steps:
       - name: Download all the dists
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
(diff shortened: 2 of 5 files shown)


.github/workflows/python-package-release-on-pypi-and-github.yml
```diff
@@ -203,6 +206,7 @@ jobs:
       - name: Publish distribution 📦 to TestPyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
         with:
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}  # zizmor: ignore[use-trusted-publishing] Optional API-token authentication.
```
dceoy, line 223: **[P1] Fail fast when the TestPyPI token is absent**

An empty `TEST_PYPI_API_TOKEN` makes `gh-action-pypi-publish` select trusted publishing, but upstream still does not support trusted publishing from an externally defined reusable workflow (`pypa/gh-action-pypi-publish#166` remains open). Therefore an external caller can enable `publish-to-testpypi-before-pypi` without this optional secret and still hit `invalid-publisher`, which then blocks the production publish. Add an explicit preflight requiring this token whenever this job runs (and apply the same contract to the default-enabled PyPI job), or limit/document the OIDC fallback to a supported same-repository case.

dceoy (commented):
The workflow fixes the Docker and distribution-name regressions, and CI passes on the current head. One publishing-authentication issue remains: external reusable-workflow callers still fall back to unsupported OIDC when optional API-token secrets are absent. This should be addressed before closing #740.

.github/workflows/python-package-release-on-pypi-and-github.yml
dceoy: Leaving this open: a hard preflight on `publish-to-testpypi` (default off) would be safe, but hard-requiring `PYPI_API_TOKEN` on `publish-to-pypi` (default-enabled) would break any external caller currently relying on tokenless OIDC. Needs a decision on whether to accept that breaking change or just document the OIDC-fallback limitation instead of enforcing it.

.github/workflows/docker-build-and-push.yml
```diff
@@ -303,7 +303,7 @@ jobs:
           while IFS= read -r tag; do
             [[ -n "${tag}" ]] || continue
             cosign verify \
-              --certificate-identity-regexp "https://github.com/${GITHUB_REPOSITORY}/.github/workflows/.+@refs/.+" \
+              --certificate-identity-regexp "https://github.com/dceoy/gh-actions-for-devops/\\.github/workflows/docker-build-and-push\\.yml@refs/.+" \
```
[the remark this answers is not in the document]
dceoy: Fixed in 99d91ea: the regex now also accepts the `@<40-char SHA>` identity form so SHA-pinned callers pass cosign verify.

.github/workflows/python-package-release-on-pypi-and-github.yml
dceoy: Documented only (66ce90b): added NOTE comments on both jobs explaining the OIDC-fallback limitation (pypa/gh-action-pypi-publish#166), without a hard preflight — avoids breaking existing tokenless-OIDC callers on the default-enabled publish-to-pypi job.

.github/workflows/python-package-release-on-pypi-and-github.yml
```diff
@@ -200,9 +203,14 @@ jobs:
...
+      # NOTE: Without TEST_PYPI_API_TOKEN, this action falls back to OIDC
+      # trusted publishing, which upstream does not yet support for an
+      # externally defined reusable workflow (pypa/gh-action-pypi-publish#166).
+      # Callers must set TEST_PYPI_API_TOKEN to publish successfully.
       - name: Publish distribution 📦 to TestPyPI
         uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
         with:
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}  # zizmor: ignore[use-trusted-publishing] Optional API-token authentication.
           repository-url: https://test.pypi.org/legacy/
   github-release:
```
dceoy, line 225: **MEDIUM — Fail fast when `TEST_PYPI_API_TOKEN` is absent**  
**Confidence: HIGH**

This secret is conditionally mandatory whenever `publish-to-testpypi-before-pypi` enables this job, but it remains optional and there is no runtime preflight. With an empty value, the upstream action treats `__token__` + empty password as Trusted Publishing and attempts the OIDC exchange; reusable workflows are still unsupported by PyPI/TestPyPI, so external callers reach the same `invalid-publisher` failure and the dependent production publish is skipped. The new NOTE documents the failure but does not satisfy issue #740's requested fail-fast behavior.

Add a step before this action that exits with an actionable error when `secrets.TEST_PYPI_API_TOKEN` is empty. Keep this check scoped to the TestPyPI job so the separate PyPI OIDC compatibility decision is unaffected.

dceoy (commented):
# PR Review Summary

1 actionable issue remains.

## Important Issues

- The TestPyPI token is conditionally required, but the workflow still invokes the publisher with an empty secret and falls into the known-unsupported reusable-workflow OIDC path. See the inline finding.

## Checked

- Docker image-reference output ordering and `GITHUB_OUTPUT` propagation
- Registry-less Docker metadata generation
- Cosign reusable-workflow identity matching, including SHA-pinned callers
- Python distribution-name normalization and version extraction
- Existing review threads and the current CI run

## Strengths

- The Docker output regression is fixed at the correct point, after target tags are added.
- The cosign identity now matches the defining reusable workflow and accepts both refs and full commit SHAs.
- Distribution filename parsing handles normalized project names without relying on a fixed hyphen field.
- Current lint and static-analysis jobs pass.

## Recommended Action

Add a TestPyPI-only preflight for the missing API token, then merge. No other high-confidence issues were identified.

.github/workflows/python-package-release-on-pypi-and-github.yml
dceoy: Fixed in 29af4bb: added a preflight step in `publish-to-testpypi` that fails fast with an actionable error when `TEST_PYPI_API_TOKEN` is empty, scoped to this job only (default-off).

PR #849, opened 2026-07-30
Make uv and pnpm versions configurable in reusable workflows

dceoy:
## Summary

- install uv and pnpm in the reusable Bats workflow
- add uv-version inputs to the Bats and Spec Kit workflows
- add pnpm-version inputs to the Bats, HTML, JSON, and TypeScript workflows
- forward pnpm-version through the JSON schema validation wrapper
- preserve latest as the default for backward compatibility
- validate caller-selected pnpm versions before npm-based installation

## Affected workflows

- .github/workflows/bats-test.yml
- .github/workflows/html-lint-and-scan.yml
- .github/workflows/json-lint.yml
- .github/workflows/json-schema-validation.yml
- .github/workflows/speckit-init.yml
- .github/workflows/typescript-package-format-and-pr.yml
- .github/workflows/typescript-package-lint-and-scan.yml
- .github/workflows/typescript-package-script.yml

## Checks

- scripts/qa.sh
- security-focused diff review (no findings)

### DIFF
--- a/.github/workflows/bats-test.yml
+++ b/.github/workflows/bats-test.yml
@@ -19,6 +19,16 @@ on:
         type: string
         description: Shell command to run after checkout, before installing Bats, to provision runtime dependencies the tests require
         default: ""
+      uv-version:
+        required: false
+        type: string
+        description: Version of uv to use
+        default: latest
+      pnpm-version:
+        required: false
+        type: string
+        description: Version of pnpm to use
+        default: latest
       runs-on:
         required: false
         type: string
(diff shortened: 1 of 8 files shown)


.github/workflows/bats-test.yml
```diff
@@ -48,6 +48,14 @@ jobs:
           printf '%s\n' "${SETUP_RUN}" > "${setup_script}"
           chmod +x "${setup_script}"
           "${setup_script}"
+      - name: Set up uv
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9  # v9.0.0
+        with:
+          version: latest
+      - name: Setup pnpm
+        uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271  # v6.0.9
```
dceoy: [P1] Do not force mutable `latest` versions in a reusable workflow. Even when a caller pins this workflow to a commit SHA, both tool binaries can change between runs; additionally, explicit `version: latest` prevents `setup-uv` from honoring `required-version` in `uv.toml`/`pyproject.toml`, and the equivalent pnpm setting bypasses the repository's `packageManager` version. This can break consumers when a new major release appears. Expose `uv-version` and `pnpm-version` workflow inputs (or otherwise honor project metadata) so callers can obtain deterministic, repository-compatible versions.

dceoy (commented):
Found one reproducibility issue in the reusable workflow. The setup actions are SHA-pinned, but the installed tool versions remain mutable and can override each consumer repository's declared versions. Please address this before merging.

PR #852, opened 2026-07-30
Add supply-chain cooldown to local QA script

dceoy:
## Summary
- Set `UV_EXCLUDE_NEWER`, `NPM_CONFIG_MIN_RELEASE_AGE`, and `PNPM_CONFIG_MINIMUM_RELEASE_AGE` in `.agents/skills/local-qa/scripts/qa.sh` so package installs during local QA skip versions published within the last 7 days.
- Reduces exposure to newly published malicious packages during local dependency resolution/installation.

## Test plan
- [ ] Run `.agents/skills/local-qa/scripts/qa.sh` and confirm it still completes successfully with the new environment variables set.

### DIFF
--- a/.agents/skills/local-qa/scripts/qa.sh
+++ b/.agents/skills/local-qa/scripts/qa.sh
@@ -6,10 +6,8 @@ REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd "${REPO_ROOT}"
 
 # Supply-chain cooldown: avoid resolving/installing packages published within the last N days.
-COOLDOWN_DAYS=7
-export UV_EXCLUDE_NEWER="${COOLDOWN_DAYS} days"
-export NPM_CONFIG_MIN_RELEASE_AGE="${COOLDOWN_DAYS}"
-export PNPM_CONFIG_MINIMUM_RELEASE_AGE=$((COOLDOWN_DAYS * 24 * 60))
+# shellcheck source=.agents/skills/local-qa/scripts/supply-chain-cooldown.sh
+source "${REPO_ROOT}/.agents/skills/local-qa/scripts/supply-chain-cooldown.sh"
 
 PYTHON_LINE_LENGTH=88
 RUFF_LINT_EXTEND_SELECT='F,E,W,C90,I,N,D,UP,S,B,A,COM,C4,PT,Q,SIM,ARG,ERA,PD,PLC,PLE,PLW,TRY,FLY,NPY,PERF,FURB,RUF'
--- a/.agents/skills/local-qa/scripts/qa.sh
+++ b/.agents/skills/local-qa/scripts/qa.sh
@@ -5,6 +5,12 @@ set -euox pipefail
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd "${REPO_ROOT}"
 
+# Supply-chain cooldown: avoid resolving/installing packages published within the last N days.
+COOLDOWN_DAYS=7
+export UV_EXCLUDE_NEWER="${COOLDOWN_DAYS} days"
+export NPM_CONFIG_MIN_RELEASE_AGE="${COOLDOWN_DAYS}"
+export PNPM_CONFIG_MINIMUM_RELEASE_AGE=$((COOLDOWN_DAYS * 24 * 60))
+
 PYTHON_LINE_LENGTH=88
 RUFF_LINT_EXTEND_SELECT='F,E,W,C90,I,N,D,UP,S,B,A,COM,C4,PT,Q,SIM,ARG,ERA,PD,PLC,PLE,PLW,TRY,FLY,NPY,PERF,FURB,RUF'
 RUFF_LINT_IGNORE='D100,D103,D203,D213,S101,B008,A002,A004,COM812,PLC2701,TRY003'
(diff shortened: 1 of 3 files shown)


dceoy (commented):
Found one blocking correctness gap. The cooldown values themselves are valid, but the current placement does not cover the skill's missing-tool installation path. Please make the policy available to those installation commands as well, then verify both in-script resolution and post-failure tool installation.

.agents/skills/local-qa/scripts/qa.sh
```diff
@@ -5,6 +5,12 @@ set -euox pipefail
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd "${REPO_ROOT}"
 
+# Supply-chain cooldown: avoid resolving/installing packages published within the last N days.
+COOLDOWN_DAYS=7
+export UV_EXCLUDE_NEWER="${COOLDOWN_DAYS} days"
+export NPM_CONFIG_MIN_RELEASE_AGE="${COOLDOWN_DAYS}"
```
dceoy: **P1 — Apply the cooldown to missing-tool installs too.** The skill runs this script as a child process, and when it fails with `command not found`, `SKILL.md` instructs the agent to install the missing tool *after the script exits*. These exports disappear with the child shell, so the subsequent `uv`/`npm`/`pnpm` installation—the highest-risk path this change is meant to protect—runs without the cooldown. Put the policy in a sourceable shared env file or explicitly require/prefix these variables in the skill's installation commands; merely exporting them inside `qa.sh` cannot affect the parent agent process.

.agents/skills/local-qa/scripts/qa.sh
```diff
@@ -5,6 +5,12 @@ set -euox pipefail
...
+export PNPM_CONFIG_MINIMUM_RELEASE_AGE=$((COOLDOWN_DAYS * 24 * 60))
```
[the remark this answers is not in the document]
dceoy: Fixed in 2c15955: moved the cooldown vars into a sourceable `scripts/supply-chain-cooldown.sh` and updated SKILL.md's install step to source it in the same command as the `npm`/`pnpm`/`uv` install.

.agents/skills/local-qa/scripts/qa.sh
dceoy: Fixed in 2c15955: cooldown vars now live in `scripts/supply-chain-cooldown.sh`, and SKILL.md's install step sources it in the same command as the install so the policy survives into the missing-tool install.

.agents/skills/local-qa/scripts/qa.sh
```diff
@@ -5,6 +5,12 @@ set -euox pipefail
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd "${REPO_ROOT}"
 
+# Supply-chain cooldown: avoid resolving/installing packages published within the last N days.
+COOLDOWN_DAYS=7
+export UV_EXCLUDE_NEWER="${COOLDOWN_DAYS} days"
```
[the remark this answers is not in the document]
dceoy: Verified on uv 0.11.23 (the version available in this environment): `--exclude-newer`/`UV_EXCLUDE_NEWER` documents and accepts duration input like "7 days" (`uv pip compile --help` lists "a duration like `3 days` or `P3D`"), and resolution succeeds with it. Duration support may postdate the 0.7.22 behavior cited here, but current tooling accepts this value, matching the same conclusion already reached on the duplicate opencode-agent finding on this line.

.agents/skills/local-qa/scripts/supply-chain-cooldown.sh
```diff
@@ -0,0 +1,9 @@
+# shellcheck shell=bash
+# Source this file (do not execute it) in the same shell command as any
+# uv/npm/pnpm install so the exported cooldown variables apply to that
+# install. Sourcing it in an earlier, separate command has no effect,
+# since exported variables do not persist across shell invocations.
+COOLDOWN_DAYS=7
+export UV_EXCLUDE_NEWER="${COOLDOWN_DAYS} days"
+export NPM_CONFIG_MIN_RELEASE_AGE="${COOLDOWN_DAYS}"
+export PNPM_CONFIG_MINIMUM_RELEASE_AGE=$((COOLDOWN_DAYS * 24 * 60))
```
dceoy, line 9: **P1 — Use pnpm’s actual generic-config environment prefix.** pnpm 11 reads generic settings from lowercase `pnpm_config_*`; `PNPM_CONFIG_MINIMUM_RELEASE_AGE` is not the generic configuration mechanism. On case-sensitive Unix shells this export is therefore ignored, so `pnpm install`/`pnpm add` runs without the intended 7-day cooldown. Export `pnpm_config_minimum_release_age` and require pnpm >=11, or version-detect and apply the appropriate pnpm-version-specific setting only around pnpm commands. Add an assertion such as `pnpm config get minimumReleaseAge` so CI verifies that the policy is effective.

dceoy (commented):
Found one blocking correctness issue in the latest head: the pnpm cooldown variable uses the wrong generic configuration prefix, so the security policy is not applied to pnpm installs on case-sensitive Unix environments. CI is green, but it does not exercise or assert this configuration behavior.

PR #858, opened 2026-07-30
Add reusable Pages, Go, and shell workflows and fix pnpm setup

### LINKED ISSUE #853: Add a deploy-only reusable workflow for GitHub Pages
## Background

The current `hugo-deploy-to-gh-pages.yml` workflow couples site build, Pages artifact upload, and deployment in one reusable workflow. Callers with repository-specific validation, tests, content generation, indexing, or output assertions need to retain their own build job while reusing the security-sensitive Pages deployment contract.

A deploy-only workflow would support those custom pipelines and remove deployment duplication from `hugo-deploy-to-gh-pages.yml`.

## Scope

- Add `.github/workflows/github-pages-deploy.yml` as a `workflow_call` reusable workflow.
- Consume an artifact previously uploaded with `actions/upload-pages-artifact`.
- Expose narrowly scoped inputs such as:
  - artifact name, defaulting to `github-pages`
  - environment name, defaulting to `github-pages`
  - runner and timeout where justified
  - bounded deployment retry count for transient failures
- Declare the minimum required permissions (`pages: write` and `id-token: write`; retain only other permissions that are demonstrably required).
- Expose the deployed `page-url` as a workflow output.
- Pin all third-party actions by commit SHA and keep versions Renovate-managed.
- Refactor `hugo-deploy-to-gh-pages.yml` into separate build/upload and deployment jobs, with the deployment job calling the new workflow.
- Add generic caller examples for:
  - a conventional Hugo site using `hugo-deploy-to-gh-pages.yml`
  - a custom build pipeline uploading a Pages artifact and calling `github-pages-deploy.yml`

## Downstream adoption

Caller-specific migrations are tracked outside this repository and are not part of this issue:

- dceoy/aims#170
- dceoy/hokf#12

## Acceptance criteria

- The deploy-only workflow successfully deploys an uploaded Pages artifact and returns its URL.
- `hugo-deploy-to-gh-pages.yml` preserves its existing caller-facing behavior and output unless an intentional breaking change is documented.
- Deployment failures are retried only up to the configured bounded limit and still fail closed.
- The custom-build example does not require arbitrary pre/post shell hooks in the reusable workflow.
- Actionlint, Zizmor, Checkov, and repository CI pass.

## Non-goals

- Modifying or migrating downstream caller repositories.
- Generalizing repository-specific Hugo validation, OKF checks, Pagefind generation, or content tests.
- Adding arbitrary command inputs or pre/post hooks to the Hugo workflow.
- Moving downstream notification policy, such as Slack alerts, into the deployment workflow.

### LINKED ISSUE #855: Add a reusable CI workflow for shell projects
## Background

The existing `shell-lint.yml` and `bats-test.yml` workflows cover ShellCheck and Bats separately, but small shell repositories often need a single coherent quality gate containing ShellCheck, shfmt verification, and Bats tests. Calling multiple reusable workflows adds jobs and setup overhead, while the current Bats workflow also prepares uv, pnpm, and Bats helper libraries even when a project does not need them.

`dceoy/oracle-pr-sentry` is a concrete candidate: its current `make check` runs ShellCheck, shfmt checks, and Bats in one minimal job.

## Scope

- Add `.github/workflows/shell-project-ci.yml` as a `workflow_call` reusable workflow.
- Run the following checks in one job:
  - ShellCheck
  - shfmt check mode
  - Bats tests
- Provide narrowly scoped inputs for:
  - shell source pathspecs or search paths
  - shfmt target paths and formatting options
  - Bats test paths
  - Bats version and optional helper-library installation
  - runner and timeout where justified
- Do not install Python, uv, Node.js, or pnpm unless explicitly required by an opt-in feature.
- Use pinned, Renovate-managed tool versions and action SHAs.
- Preserve fail-fast behavior and clear per-check log grouping.
- Add execution coverage using repository fixtures for a passing project and representative ShellCheck, shfmt, and Bats failures.
- Document a migration example matching the behavior of `dceoy/oracle-pr-sentry`'s current `make check` job.

## Acceptance criteria

- A caller can reproduce ShellCheck + shfmt + Bats validation with one reusable-workflow job.
- Default execution installs only the tools required for those checks.
- Paths containing spaces and empty file selections are handled explicitly and safely.
- A formatting violation, lint violation, or failed Bats test causes the job to fail.
- Existing `shell-lint.yml` and `bats-test.yml` remain focused building blocks; the new workflow does not silently change their contracts.
- Actionlint, ShellCheck, shfmt, Bats, Zizmor, Checkov, and repository CI pass.

## Non-goals

- Supporting arbitrary setup shell commands as a generic escape hatch.
- Adding language runtimes unrelated to the tested shell project by default.
- Replacing application-specific integration or end-to-end test environments.

### LINKED ISSUE #856: Add a reusable Go package test workflow
## Background

`go-package-lint-and-scan.yml` covers golangci-lint, gosec, govulncheck, and `go mod tidy` verification, but it does not provide the standard correctness checks commonly required by Go repositories: gofmt verification, `go vet`, race-enabled tests, coverage generation, and build verification.

Repositories such as `dceoy/segh` currently implement these generic checks locally alongside repository-specific configuration validation. A separate test workflow would allow callers to reuse the generic Go quality gate without expanding the security-scanning workflow into a monolithic interface.

## Scope

- Add `.github/workflows/go-package-test.yml` as a `workflow_call` reusable workflow.
- Support the following checks with safe defaults:
  - gofmt verification
  - `go vet`
  - `go test`, optionally with `-race`
  - coverage profile generation and optional artifact upload
  - `go build` verification
- Provide narrowly scoped inputs for:
  - Go version
  - working directory
  - package pattern, defaulting to `./...`
  - race detector enablement
  - coverage enablement, profile path, and artifact retention
  - build enablement and package pattern
  - runner and timeout where justified
- Use `actions/setup-go` caching with lockfile-sensitive behavior.
- Pin all actions by commit SHA and keep versions Renovate-managed.
- Add execution tests or smoke fixtures covering success, formatting failure, test failure, race/coverage argument construction, and nested working directories.
- Document how callers can combine this workflow with `go-package-lint-and-scan.yml` and retain repository-specific validation in separate local jobs.

## Acceptance criteria

- A caller can run formatting, vetting, tests, coverage, and build verification without arbitrary shell-command inputs.
- Race-enabled tests and coverage can be independently enabled or disabled.
- Coverage artifacts are uploaded only when requested and are named deterministically.
- `go test` and `go build` package patterns are validated and safely passed to Go tooling.
- The workflow does not absorb repository-specific configuration validation or application setup.
- Existing `go-package-lint-and-scan.yml` remains focused on linting and vulnerability/security analysis.
- Actionlint, Zizmor, Checkov, and repository CI pass.

## Non-goals

- Replacing repository-specific checks such as `segh` configuration/schema validation.
- Adding arbitrary pre-test or post-test command hooks.
- Combining all Go linting, security scanning, testing, release, and deployment behavior into one workflow.

### LINKED ISSUE #857: Avoid conflicting pnpm versions when package.json defines packageManager
## Problem

Reusable workflows that use `pnpm/action-setup` currently define `pnpm-version` with the default value `latest` and always pass it as the action's `version` input.

This conflicts with repositories that already pin pnpm in `package.json`, for example:

```json
{
  "packageManager": "pnpm@11.16.0"
}
```

`pnpm/action-setup` rejects the two differing sources:

```text
Error: Multiple versions of pnpm specified:
  - version latest in the GitHub Action config with the key "version"
  - version pnpm@11.16.0 in the package.json with the key "packageManager"
Remove one of these versions to avoid version mismatch errors like ERR_PNPM_BAD_PM_VERSION
```

Observed failure:

- https://github.com/dceoy/dceoy.github.io/actions/runs/30540949407/job/90865327329?pr=93

The regression was introduced when reusable workflows switched from globally installing pnpm to `pnpm/action-setup` while retaining `latest` as the default override.

## Proposed change

Treat `package.json#packageManager` as the default source of truth:

1. Change the `pnpm-version` input default from `latest` to an empty string.
2. Pass the empty value to `pnpm/action-setup`, allowing it to infer the version from `packageManager` or `devEngines.packageManager`.
3. Set `package_json_file` relative to the repository root when `package-path` is configurable.
4. Preserve `pnpm-version` as an explicit override for callers that do not pin pnpm in `package.json`.
5. Apply the behavior consistently to every reusable workflow using `pnpm/action-setup`.

Example:

```yaml
pnpm-version:
  required: false
  type: string
  description: Optional pnpm version override
  default: ""

# ...

- name: Setup pnpm
  if: env.PACKAGE_MANAGER == 'pnpm'
  uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271
  with:
    version: ${{ inputs.pnpm-version }}
    package_json_file: ${{ format('{0}/package.json', inputs.package-path) }}
```

## Acceptance criteria

- A caller with `packageManager: pnpm@<version>` and no `pnpm-version` input succeeds.
- An explicit matching `pnpm-version` input succeeds.
- An explicit conflicting version continues to fail clearly rather than silently overriding the repository pin.
- pnpm projects located under a non-root `package-path` resolve the correct `package.json`.
- npm-based workflows remain unchanged.
- Regression coverage is added for root and nested pnpm projects.

dceoy:
## Summary

- add a deploy-only GitHub Pages reusable workflow with bounded fail-closed retries, minimal permissions, artifact selection, and a `page-url` output
- refactor the Hugo Pages workflow into build/upload and reusable deployment jobs without changing its caller-facing output
- add focused reusable workflows for Go correctness checks and combined shell-project CI
- make `pnpm/action-setup` infer pnpm from each project's `package.json` unless callers provide an explicit override
- add smoke fixtures and Bats coverage for Go, shell, and pnpm root/nested behavior, and document caller examples

## Why

Repositories need reusable deployment and quality-gate contracts without repository-specific hooks or monolithic workflows. pnpm callers also need their checked-in `packageManager` pin to remain the default source of truth instead of conflicting with a forced `latest` override.

## Impact

Callers can deploy a prebuilt Pages artifact, run standard Go formatting/vet/test/coverage/build checks, or run ShellCheck/shfmt/Bats in one job. Existing focused workflows remain available, and the Hugo workflow preserves its existing inputs and `page-url` output. pnpm callers without a repository pin can continue to set `pnpm-version` explicitly.

## Root cause

The existing Pages workflow coupled build and deployment, no reusable workflow covered the standard Go correctness gate or minimal combined shell gate, and pnpm workflows always passed `latest` even when `package.json` declared a different package-manager version.

## Validation

- `scripts/qa.sh` (shfmt, ShellCheck, Prettier, Zizmor, Actionlint, yamllint, Checkov)
- `bats --print-output-on-failure .github/scripts/tests/*.bats` (16 tests)
- README regeneration with `.github/scripts/update-readme.sh`
- security-focused diff review; no actionable findings

Closes #853
Closes #855
Closes #856
Closes #857


dceoy (commented):
Found two merge-blocking issues:

1. The repository CI is currently red because the new shell workflow Bats coverage executes the extracted shfmt step without provisioning `shfmt` in `readme-generator-test`.
2. The Go coverage artifact name is fixed to `go-coverage`, so multiple invocations or matrix entries in one workflow run will conflict under `actions/upload-artifact@v7`.

The Pages workflow split, bounded fail-closed retry flow, and pnpm packageManager inference otherwise look coherent. Please address the inline findings and rerun CI.

.github/workflows/ci.yml
```diff
@@ -65,10 +65,15 @@ jobs:
...
+        with:
+          go-version: stable
+          cache: false
       - name: Setup Bats and Bats libraries
         uses: bats-core/bats-action@77d6fb60505b4d0d1d73e48bd035b55074bbfb43  # 4.0.0
       - name: Run Bats tests
-        run: bats --print-output-on-failure .github/scripts/tests/update-readme.bats
+        run: bats --print-output-on-failure .github/scripts/tests/*.bats
   github-actions-lint-and-scan:
     if: >
```
dceoy, line 78: [P1] Provision `shfmt` before running the new workflow-script tests. This job now executes `shell-project-ci.bats`, which extracts and runs the `Check formatting with shfmt` script, but none of the preceding steps installs `shfmt`. The current CI run fails with `shfmt: command not found`, so repository CI cannot pass as submitted. Install the pinned shfmt version here, or make the test harness provision it explicitly.

.github/workflows/go-package-test.yml
```diff
@@ -0,0 +1,178 @@
...
+          fi
+          test_args+=("${GO_TEST_PACKAGE_PATTERN}")
+          go "${test_args[@]}"
+      - name: Upload coverage profile
+        if: inputs.coverage-enabled && inputs.upload-coverage
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
+        with:
+          name: go-coverage
+          path: ${{ inputs.working-directory }}/${{ inputs.coverage-profile }}
+          if-no-files-found: error
```
dceoy, line 196: [P1] Make the coverage artifact name configurable or otherwise unique per invocation. Every call currently uploads `go-coverage`; `actions/upload-artifact@v4+` artifacts are immutable and duplicate names across jobs/matrix entries in the same workflow run fail with a conflict. A caller testing multiple Go modules with `upload-coverage: true` will therefore break. Add a deterministic `coverage-artifact-name` input (or an equivalent unique naming contract) and cover the multi-invocation case.

.github/workflows/hugo-deploy-to-gh-pages.yml
```diff
@@ -78,26 +78,18 @@ on:
...
-concurrency:
-  group: pages
-  cancel-in-progress: false
 defaults:
   run:
     shell: bash -euo pipefail {0}
     working-directory: .
 jobs:
-  build-and-deploy:
+  build:
```
dceoy, line 94: [P1] Preserve concurrency across the Hugo build as well as deployment. Moving the static `pages` concurrency group into the deploy-only reusable workflow allows separate Hugo runs to build concurrently. If an older run builds more slowly, a newer run can deploy first and the older run can then enter the deploy group later and overwrite Pages with stale content; with GitHub concurrency's single pending slot, arrival at the deploy job can also replace a newer pending deployment. Keep workflow-level concurrency in this combined build-and-deploy wrapper (while retaining deployment-level concurrency for deploy-only callers), so deployment order is based on the workflow runs rather than build completion order.

dceoy (commented):
Found one merge-blocking deployment-order regression. The earlier CI/tooling and coverage-artifact findings are resolved, all review threads are resolved, and the latest CI run passes. Please restore whole-workflow concurrency for the combined Hugo workflow so an older build cannot deploy after a newer one.

.github/workflows/hugo-deploy-to-gh-pages.yml
```diff
@@ -142,7 +137,16 @@ jobs:
...
-      - name: Deploy to GitHub Pages
-        id: deployment
-        uses: actions/deploy-pages@cd2ce8fcbc39b97be8ca5fce6e763baed58fa128  # v5.0.0
+  deploy:
+    needs:
+      - build
+    permissions:
+      id-token: write
+      pages: write
+    uses: ./.github/workflows/github-pages-deploy.yml
```
dceoy, line 150: [P1] Avoid re-entering the same `pages` concurrency group from the nested reusable workflow. This wrapper still owns workflow-level `concurrency.group: pages`, while `github-pages-deploy.yml` also declares the same workflow-level group. The called deployment workflow therefore competes with its own parent run and can remain pending or be rejected as a concurrency deadlock before `deploy-pages` executes. Keep the wrapper-level serialization needed to prevent stale builds, but use distinct group names (for example `hugo-pages` for the wrapper and `pages-deploy` for deploy-only serialization), or make the callee group explicitly overridable.

.github/workflows/hugo-deploy-to-gh-pages.yml
```diff
@@ -91,13 +91,8 @@ defaults:
     shell: bash -euo pipefail {0}
     working-directory: .
 jobs:
-  build-and-deploy:
+  build:
```
dceoy, line 94: [P1] Restrict the new build job to `contents: read`. After splitting build and deployment, this job still inherits the workflow-level `pages: write` and `id-token: write` permissions, so every build step—including the third-party Hugo setup action—runs in a deployment-privileged job. Add `permissions: { contents: read }` to `build` (or move all permissions to the individual jobs) and keep `pages: write`/`id-token: write` only on `deploy`.

.github/workflows/go-package-test.yml
```diff
@@ -0,0 +1,212 @@
...
+          fi
+          if [[ ! -d "${GO_WORKING_DIRECTORY}" ]]; then
+            echo "::error::working-directory does not exist: ${GO_WORKING_DIRECTORY}"
+            exit 1
+          fi
+      - name: Set up Go
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
+        with:
+          go-version: ${{ inputs.go-version }}
+          cache: ${{ hashFiles(format('{0}/go.sum', inputs.working-directory), format('{0}/go.work.sum', inputs.working-directory)) != '' }}
```
dceoy: [P2] Include nested module sums in the workspace cache key. For a conventional `go.work` root, dependency files normally live at `module-a/go.sum`, `module-b/go.sum`, etc.; this expression checks only root `go.sum`/`go.work.sum`, so caching is disabled when `go.work.sum` is absent and ignores module dependency changes when it exists. Use a workspace-relative glob such as `${{ inputs.working-directory }}/**/go.sum` together with `go.work.sum` in both the enablement hash and `cache-dependency-path`, and cover it with the workspace fixture.

dceoy (commented):
Found three issues in the current head. The concurrency collision is merge-blocking because the combined Hugo workflow can deadlock before deployment; the build job also retains deployment credentials after the split. The Go workspace cache configuration should include nested module lockfiles. Repository CI is green, but it does not execute a real Pages deployment and therefore cannot catch the first two issues.

.github/workflows/hugo-deploy-to-gh-pages.yml
dceoy: Renamed the wrapper's concurrency group to `hugo-pages` (9ed7235); `github-pages-deploy.yml` keeps `pages` and is the only caller, so no other collision exists.

.github/workflows/hugo-deploy-to-gh-pages.yml
dceoy: Restricted `build` to `permissions: contents: read` (9ed7235); `pages: write`/`id-token: write` now apply to `deploy` only.

.github/workflows/go-package-test.yml
dceoy: Switched to a `**/go.sum` glob for both the cache-enablement hash and `cache-dependency-path` (9ed7235), and added a bats test on the workspace fixture that fails against the old root-only pattern and passes with the fix.

dceoy (commented):
Reviewed the current head (`9ed7235`). No further actionable findings.

The previously identified issues around tool provisioning, coverage artifact naming, Go workspace support/cache coverage, Hugo deployment concurrency, and job-level permissions are resolved. The latest CI/CD run is green.

PR #860, opened 2026-07-30
Avoid conflicting pnpm version sources

### LINKED ISSUE #857: Avoid conflicting pnpm versions when package.json defines packageManager

dceoy:
## Summary

- default `pnpm-version` to an empty string so `pnpm/action-setup` can use `package.json#packageManager`
- pass the root or `package-path`-relative `package.json` to every direct `pnpm/action-setup` invocation
- preserve explicit version inputs and add regression coverage for root, nested, and wrapper workflows
- run all Bats test files in CI

## Root cause

The reusable workflows always passed the default `latest` as the action's `version` input. Repositories that pin pnpm in `package.json` therefore supplied two conflicting version sources, which `pnpm/action-setup` correctly rejected.

## Validation

- Bats: 7 tests passed
- `shfmt` and ShellCheck passed
- Actionlint and Yamllint passed for all workflows
- Zizmor reported no findings
- Checkov passed in offline mode with `--skip-download --framework github_actions`
- `git diff --check` passed

Closes #857

### DIFF
--- a/.github/scripts/tests/pnpm-setup.bats
+++ b/.github/scripts/tests/pnpm-setup.bats
@@ -10,31 +10,40 @@ setup() {
 pnpm_setup_value() {
   local workflow="$1" key="$2"
   yq -r \
-    ".jobs.*.steps[] | select(.uses | startswith(\"pnpm/action-setup@\")) | .with.${key}" \
+    ".jobs.*.steps[] | select(.uses | test(\"^pnpm/action-setup@\")) | .with.${key}" \
     "${workflow}"
 }
 
 @test "pnpm version defaults to package.json in every exposed input" {
+  local count=0
   while IFS= read -r workflow; do
     run yq -r '.on.workflow_call.inputs.pnpm-version.default' "${workflow}"
     [ "${status}" -eq 0 ]
     [ "${output}" = "" ]
+    count=$((count + 1))
   done < <(grep -rl --include='*.yml' '^      pnpm-version:$' "${WORKFLOWS}")
+  [ "${count}" -gt 0 ]
 }
 
 @test "explicit pnpm version remains an action override" {
+  local count=0
   while IFS= read -r workflow; do
     run pnpm_setup_value "${workflow}" version
     [ "${status}" -eq 0 ]
     [ "${output}" = "\${{ inputs.pnpm-version || env.PNPM_VERSION }}" ]
+    count=$((count + 1))
   done < <(grep -rl --include='*.yml' 'pnpm/action-setup@' "${WORKFLOWS}")
+  [ "${count}" -gt 0 ]
 }
 
 @test "pnpm setup preserves latest fallback without package metadata" {
+  local count=0
   while IFS= read -r workflow; do
     run grep -F "echo 'PNPM_VERSION=latest'" "${workflow}"
     [ "${status}" -eq 0 ]
+    count=$((count + 1))
   done < <(grep -rl --include='*.yml' 'pnpm/action-setup@' "${WORKFLOWS}")
+  [ "${count}" -gt 0 ]
 }
 
 @test "root pnpm project resolves the root package.json" {
@@ -44,10 +53,13 @@ pnpm_setup_value() {
 }
 
 @test "nested pnpm projects resolve package.json from package-path" {
+  local count=0
   while IFS= read -r workflow; do
     [[ "${workflow}" == */bats-test.yml ]] && continue
     run pnpm_setup_value "${workflow}" package_json_file
     [ "${status}" -eq 0 ]
     [ "${output}" = "\${{ format('{0}/package.json', inputs.package-path) }}" ]
+    count=$((count + 1))
   done < <(grep -rl --include='*.yml' 'pnpm/action-setup@' "${WORKFLOWS}")
+  [ "${count}" -gt 0 ]
 }
--- a/.github/scripts/tests/pnpm-setup.bats
+++ b/.github/scripts/tests/pnpm-setup.bats
@@ -1,0 +1,53 @@
+#!/usr/bin/env bats
+
+bats_require_minimum_version 1.5.0
+
+setup() {
+  REPO_ROOT="$(cd "${BATS_TEST_DIRNAME}/../../.." && pwd)"
+  WORKFLOWS="${REPO_ROOT}/.github/workflows"
+}
+
+pnpm_setup_value() {
+  local workflow="$1" key="$2"
+  yq -r \
+    ".jobs.*.steps[] | select(.uses | startswith(\"pnpm/action-setup@\")) | .with.${key}" \
+    "${workflow}"
+}
+
+@test "pnpm version defaults to package.json in every exposed input" {
+  while IFS= read -r workflow; do
+    run yq -r '.on.workflow_call.inputs.pnpm-version.default' "${workflow}"
+    [ "${status}" -eq 0 ]
+    [ "${output}" = "" ]
+  done < <(grep -rl --include='*.yml' '^      pnpm-version:$' "${WORKFLOWS}")
+}
+
+@test "explicit pnpm version remains an action override" {
+  while IFS= read -r workflow; do
+    run pnpm_setup_value "${workflow}" version
+    [ "${status}" -eq 0 ]
+    [ "${output}" = "\${{ inputs.pnpm-version || env.PNPM_VERSION }}" ]
+  done < <(grep -rl --include='*.yml' 'pnpm/action-setup@' "${WORKFLOWS}")
+}
+
+@test "pnpm setup preserves latest fallback without package metadata" {
+  while IFS= read -r workflow; do
+    run grep -F "echo 'PNPM_VERSION=latest'" "${workflow}"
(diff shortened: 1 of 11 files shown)


.github/scripts/tests/pnpm-setup.bats
```diff
@@ -0,0 +1,53 @@
...
+
+setup() {
+  REPO_ROOT="$(cd "${BATS_TEST_DIRNAME}/../../.." && pwd)"
+  WORKFLOWS="${REPO_ROOT}/.github/workflows"
+}
+
+pnpm_setup_value() {
+  local workflow="$1" key="$2"
+  yq -r \
+    ".jobs.*.steps[] | select(.uses | startswith(\"pnpm/action-setup@\")) | .with.${key}" \
```
dceoy: `startswith(...)` is not accepted by the repository's installed `yq`; the current CI fails with `lexer: invalid input text "startswith(\"pnpm..."` in three tests. Reuse the selector already used by `pnpm-version-inference.bats`, e.g. `select(.uses | test("^pnpm/action-setup@"))`, so digest pin updates remain tolerated without breaking the test parser.

dceoy (commented):
Two blocking issues remain on the current head:

1. `.github/workflows/json-schema-validation.yml` still defaults `pnpm-version` to `latest` and forwards that value to `json-lint.yml`. Callers using this wrapper therefore still provide an explicit `latest` alongside `package.json#packageManager`, reproducing the original conflict. Change the wrapper input default to `""`, update its description, and cover the wrapper path in the regression tests.

2. CI is currently failing in `readme-generator-test`. Besides the inline `yq` selector failure, the pre-existing `.github/scripts/tests/pnpm-version-inference.bats` still expects each action input to be exactly `${{ inputs.pnpm-version }}`. Update that test to reflect the intentional fallback expression (or consolidate the duplicate pnpm tests) so the full Bats suite passes.

The workflow linters and security scans pass, but this PR should not merge until the Bats suite is green and the wrapper path no longer preserves the original bug.

.github/workflows/bats-test.yml
```diff
@@ -27,8 +27,8 @@ on:
       pnpm-version:
         required: false
         type: string
-        description: Version of pnpm to use
-        default: latest
+        description: Optional pnpm version override
+        default: ""
```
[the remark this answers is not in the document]
dceoy: Fixed: each affected workflow now falls back to `latest` via `has_pnpm_version_source` when `packageManager` is absent, and `json-schema-validation.yml`'s own default was updated to `""` so its wrapper call to `json-lint.yml` benefits from that fallback too.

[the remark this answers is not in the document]
dceoy: Since the fallback preserves the old `latest` behavior when `packageManager` is absent, there's no new hard requirement to document here.

.github/scripts/tests/pnpm-setup.bats
```diff
@@ -0,0 +1,46 @@
...
+
+setup() {
+  REPO_ROOT="$(cd "${BATS_TEST_DIRNAME}/../../.." && pwd)"
+  WORKFLOWS="${REPO_ROOT}/.github/workflows"
+}
+
+pnpm_setup_value() {
+  local workflow="$1" key="$2"
+  yq -r \
+    ".jobs.*.steps[] | select(.uses == \"pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271\") | .with.${key}" \
```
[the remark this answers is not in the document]
dceoy: Fixed using `test("^pnpm/action-setup@")` (matching `pnpm-version-inference.bats`) instead of `startswith`, which isn't a valid yq function here and was breaking CI.

.github/scripts/tests/pnpm-setup.bats
dceoy: Fixed in f4e5669 exactly as suggested.

dceoy:
Items 1 (SHA hardcoding) and 2 (vacuous grep pass) fixed in f4e5669. Item 3 (ci.yml wildcard) predates this PR (from #858) and isn't part of this diff, so it's out of scope here.

.github/workflows/html-lint-and-scan.yml
```diff
@@ -99,15 +99,25 @@ jobs:
...
           persist-credentials: false
       - name: Detect the package manager
+        env:
+          PNPM_VERSION: ${{ inputs.pnpm-version }}
         working-directory: ${{ inputs.package-path }}
         run: |
+          has_pnpm_version_source() {
+            local manifest
+            manifest="$(tr -d '\n\r' < package.json 2>/dev/null || true)"
+            grep -qE '"packageManager"[[:space:]]*:[[:space:]]*"pnpm@[^"]+"|"devEngines"[[:space:]]*:[[:space:]]*\{[^}]*"packageManager"[[:space:]]*:[[:space:]]*\{[^}]*("name"[[:space:]]*:[[:space:]]*"pnpm"[^}]*"version"[[:space:]]*:[[:space:]]*"[^"]+"|"version"[[:space:]]*:[[:space:]]*"[^"]+"[^}]*"name"[[:space:]]*:[[:space:]]*"pnpm")' <<< "${manifest}"
```
dceoy: **[P1] Parse `devEngines` structurally before choosing `latest`**

This regex cannot cross a closing `}` before `packageManager`. For example, the valid manifest below is supported by `pnpm/action-setup`, but `has_pnpm_version_source` returns false because the nested `runtime` object appears first:

```json
{"devEngines":{"runtime":{"name":"node","version":">=22"},"packageManager":{"name":"pnpm","version":"10.33.0"}}}
```

The step therefore exports `PNPM_VERSION=latest`; because an explicit `version` is then supplied, `pnpm/action-setup` silently uses `latest` instead of the declared `devEngines.packageManager` pin. The same detector is duplicated across the other affected workflows. Please parse `package.json` as JSON (for example with the runner's Node.js runtime) and add a regression case where another nested `devEngines` property precedes `packageManager`.

dceoy (commented):
The default-version conflict and the earlier review findings are addressed, and the latest CI run is green. One correctness issue remains in the fallback detector: it can miss valid `devEngines.packageManager` declarations and silently install `latest` instead of the repository pin. Please parse the manifest structurally and add coverage for a `devEngines` object containing another nested property before `packageManager`.

.github/workflows/html-lint-and-scan.yml
dceoy: Fixed in 1cc78af: `has_pnpm_version_source` now parses `package.json` structurally with `jq` (navigating by key, not sequential regex matching) instead of a brace-balanced regex, so a nested `devEngines` property preceding `packageManager` no longer breaks detection. Applied the same fix to the other 4 duplicates and to `bats-test.yml`'s inline copy (now wrapped in the same function for consistency), and added regression fixtures/tests for this case and the no-source case. Went with `jq` rather than Node.js since it's the established idiom across this repo's workflows and Node isn't set up yet at this step.

PR #861, opened 2026-07-31
Add create-generated-update-pr action and command-based shell CI

### LINKED ISSUE #854: Add a composite action for generated-file update pull requests
## Background

Scheduled data or documentation update workflows commonly repeat the same Git operations after generating files: detect changes, create or update a branch, commit a restricted set of generated paths, push, and open or update a pull request.

This logic should be a composite action rather than a reusable workflow because it must run as a step in the same job and workspace after repository-specific generation and validation steps.

## Scope

- Add a composite action under `.github/actions/create-generated-update-pr/`.
- Provide an idempotent flow that:
  - exits successfully with explicit outputs when no scoped files changed
  - stages only caller-supplied pathspecs; never use an unscoped `git add .`
  - creates or resets a deterministic update branch
  - commits and pushes generated changes
  - creates a new pull request or updates the existing open pull request for that branch
  - optionally applies caller-supplied labels and draft state
- Support narrowly scoped inputs such as:
  - required changed-file pathspecs
  - branch name
  - base branch
  - commit message
  - pull request title and body
  - labels and draft mode
  - Git author name and email with safe bot defaults
- Accept authentication through `GH_TOKEN` in the step environment rather than a token action input.
- Expose outputs including `changed`, `branch`, `commit-sha`, `pr-number`, and `pr-url`.
- Extract non-trivial shell logic into testable scripts and add Bats coverage for no-change, new-PR, existing-PR, restricted-staging, and failure paths.
- Document the minimum caller permissions (`contents: write` and `pull-requests: write`).
- Add a generic usage example that runs after caller-owned generation and validation steps.

## Downstream adoption

Caller-specific migration is tracked outside this repository and is not part of this issue:

- dceoy/aims#171

## Acceptance criteria

- Re-running the action updates the same branch and pull request instead of creating duplicates.
- Unrelated workspace changes are never staged or committed.
- No-change runs are successful and produce `changed=false` without pushing a branch.
- Inputs are safely quoted and validated; branch names and pathspecs cannot cause command injection.
- The implementation works after arbitrary caller generation and validation steps in the same job.
- Actionlint, ShellCheck, shfmt, Bats, Zizmor, Checkov, and repository CI pass.

## Non-goals

- Modifying or migrating downstream caller repositories.
- Running arbitrary generation commands inside the action.
- Replacing repository-specific validation, change summaries, or notification logic.
- Automatically merging the generated pull request.

### LINKED ISSUE #859: Add a reusable CI workflow for shell-only projects
## Summary

Add a reusable GitHub Actions workflow for small shell-only repositories so callers do not need to duplicate tool installation, version pinning, and checksum logic.

The initial consumer is `dceoy/oracle-pr-sentry`, whose CI currently installs and runs Bats, ShellCheck, shfmt, zizmor, actionlint, yamllint, and Checkov locally.

## Scope

- Add a reusable workflow for shell projects.
- Install and cache the required QA tools centrally.
- Pin tool versions in this repository.
- Accept a caller-provided QA command, such as:

  ```text
  .agents/skills/local-qa/scripts/qa.sh
  ```

- Preserve least-privilege default permissions.
- Support Ubuntu 24.04 callers.
- Keep the interface intentionally small; do not expose every underlying tool option.

## Suggested interface

```yaml
jobs:
  ci:
    uses: dceoy/gh-actions-for-devops/.github/workflows/shell-project-ci.yml@<ref>
    with:
      command: .agents/skills/local-qa/scripts/qa.sh
```

Optional inputs should be added only when multiple real consumers require them.

## Acceptance criteria

- A reusable workflow runs Bats, ShellCheck, shfmt, zizmor, actionlint, yamllint, and Checkov through the caller's QA command.
- Tool installation and version pins no longer need to be copied into the caller workflow.
- The workflow uses read-only repository permissions by default.
- `oracle-pr-sentry` can migrate without losing any current CI checks.
- Documentation includes a minimal caller example.

## Non-goals

- Supporting arbitrary build systems or language package managers.
- Replacing repository-specific QA scripts.
- Adding a broad matrix or configuration surface before another consumer needs it.

dceoy:
## Summary

- Add a `create-generated-update-pr` composite action (`.github/actions/create-generated-update-pr/`) that stages caller-scoped generated file changes, commits and force-pushes them to a deterministic branch, and creates or updates that branch's pull request via `gh`, authenticating through `GH_TOKEN` in the step environment (not an action input). Non-trivial logic lives in testable scripts under `scripts/`.
- Redesign `shell-project-ci.yml` around a required `command` input: install ShellCheck, shfmt, actionlint, Bats, zizmor, yamllint, and Checkov centrally with pinned versions, then run the caller's QA command (e.g. `.agents/skills/local-qa/scripts/qa.sh`).
- Extend zizmor coverage (`qa.sh` and `ci.yml`) to `.github/actions` now that a composite action exists there.
- Document both features in `README.md.tmpl` and regenerate `README.md`.

## Breaking change note

`shell-project-ci.yml`'s interface changes from the granular `shell-file-names`/`shfmt-*`/`bats-test-paths` inputs (added one day earlier in #858 for #855) to a single `command` input. #859 explicitly asks for a caller-provided QA command that also installs zizmor/actionlint/yamllint/Checkov, which conflicts with #855's "no arbitrary command escape hatch" non-goal — since #859 is a distinct, later, still-open issue targeting the same file, this PR replaces the just-shipped granular interface rather than adding a second mode. Any caller who has already adopted last night's interface will need to switch to `command:`.

## Why

- #854: scheduled generation/documentation workflows repeatedly reimplement the same detect-changes/branch/commit/push/PR-upsert logic; a composite action lets callers reuse it as a step after their own generation/validation, in the same job/workspace.
- #859: `oracle-pr-sentry` and similar shell-only repos currently install and pin ShellCheck, shfmt, Bats, zizmor, actionlint, yamllint, and Checkov locally; centralizing installation while letting the caller keep its own QA script avoids duplicating tool installation and version pinning.

## Validation

- `scripts/qa.sh` (shfmt, ShellCheck, Prettier, zizmor, actionlint, yamllint, Checkov) — clean
- `bats --print-output-on-failure .github/scripts/tests/*.bats` — 41 tests pass, including:
  - `create-generated-update-pr.bats`: no-change, restricted-staging (unrelated files never staged), new-PR, existing-PR-update, invalid/flag-like branch names, and missing-`GH_TOKEN` failure paths
  - `shell-project-ci.bats`: input validation, successful/failing/multi-line caller commands, and working-directory propagation
- README regenerated with `.github/scripts/update-readme.sh` + Prettier
- Verified executable bits on the new action scripts (`100755`) and that `git ls-files` picks up all new shell files for ShellCheck/shfmt
- Manually confirmed `gh auth setup-git` (stubbed in tests) is the auth path for `git push`, independent of `persist-credentials`

Closes #854
Closes #859


.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
```diff
@@ -0,0 +1,25 @@
...
+
+if [[ "${BRANCH}" == -* ]] || ! git check-ref-format --branch "${BRANCH}" > /dev/null 2>&1; then
+  echo "::error::branch is not a valid Git branch name: ${BRANCH}" >&2
+  exit 1
+fi
+
+gh auth setup-git
+
+git checkout -B "${BRANCH}"
+git -c user.name="${GIT_USER_NAME}" -c user.email="${GIT_USER_EMAIL}" commit --message "${COMMIT_MESSAGE}"
```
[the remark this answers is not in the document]
dceoy: Fixed in 2ea51a0: `commit-and-push.sh` now receives `PATHS` and commits with `--only -- ${pathspecs[@]}`, so unrelated staged entries from earlier steps are left uncommitted.

.github/workflows/shell-project-ci.yml
```diff
@@ -205,21 +190,11 @@ jobs:
...
-          files=()
-          while IFS= read -r -d '' file; do
-            files+=("./${file}")
-          done < <(git ls-files -z -- "${pathspecs[@]}")
-          [[ ${#files[@]} -gt 0 ]] || { echo "::error::Bats selected no tracked test files"; exit 1; }
-          bats "${files[@]}"
-          echo "::endgroup::"
+          qa_command_script="$(mktemp)"
+          printf '%s\n' "${QA_COMMAND}" > "${qa_command_script}"
+          bash -euo pipefail "${qa_command_script}"
```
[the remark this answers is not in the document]
dceoy: Fixed in 2ea51a0: the QA step now fails with `::error::` if `git diff --quiet -- .` finds tracked-file changes after the command runs, catching write-mode formatters like qa.sh's `shfmt --write`.

.github/workflows/shell-project-ci.yml
```diff
@@ -153,50 +144,44 @@ jobs:
...
+          if command -v actionlint >/dev/null; then
+            exit 0
+          fi
+          [[ "${RUNNER_OS}" == "Linux" ]] || { echo "::error::automatic actionlint installation requires Linux"; exit 1; }
+          arch=amd64
+          [[ "${RUNNER_ARCH}" == "ARM64" ]] && arch=arm64
+          mkdir -p "${HOME}/.local/bin"
+          actionlint_version="${ACTIONLINT_VERSION#v}"
+          curl --fail --location --silent --show-error \
+            "https://github.com/rhysd/actionlint/releases/download/${ACTIONLINT_VERSION}/actionlint_${actionlint_version}_linux_${arch}.tar.gz" \
```
dceoy: **[P1] Preserve actionlint checksum verification**

This streams the release archive directly into `tar`, so migrating `oracle-pr-sentry` to this workflow removes its current download of `actionlint_*_checksums.txt` and `sha256sum` verification. That is a supply-chain regression and conflicts with #859's goal of centralizing the caller's checksum logic. Download the archive and published checksum file, verify the selected asset, and only then extract/install it.

.github/workflows/shell-project-ci.yml
```diff
@@ -205,21 +190,16 @@ jobs:
...
-          while IFS= read -r -d '' file; do
-            files+=("./${file}")
-          done < <(git ls-files -z -- "${pathspecs[@]}")
-          [[ ${#files[@]} -gt 0 ]] || { echo "::error::Bats selected no tracked test files"; exit 1; }
-          bats "${files[@]}"
-          echo "::endgroup::"
+          qa_command_script="$(mktemp)"
+          printf '%s\n' "${QA_COMMAND}" > "${qa_command_script}"
+          bash -euo pipefail "${qa_command_script}"
+          if ! git diff --quiet -- .; then
```
dceoy: **[P2] Check the whole repository and the index for QA rewrites**

Because this step runs under `search-path`, `git diff --quiet -- .` only examines unstaged changes below that directory. The documented `qa.sh` changes to the repository root and can rewrite tracked files outside `search-path`; a command that stages its rewrites is also invisible to this check. Run the cleanliness check from `GITHUB_WORKSPACE` and include both index and worktree changes (for example, a root-level `git status --porcelain --untracked-files=no`).

.github/actions/create-generated-update-pr/scripts/create-or-update-pr.sh
```diff
@@ -0,0 +1,48 @@
...
+if [[ -n "${LABELS:-}" ]]; then
+  while IFS= read -r label; do
+    [[ -n "${label}" ]] && labels+=("${label}")
+  done < <(tr ',' '\n' <<< "${LABELS}")
+fi
+
+existing_pr_number="$(gh pr list --head "${BRANCH}" --base "${BASE}" --state open --json number --jq '.[0].number // empty')"
+
+if [[ -n "${existing_pr_number}" ]]; then
+  gh pr edit "${existing_pr_number}" --title "${TITLE}" --body "${BODY:-}"
```
dceoy: **[P2] Apply base and draft inputs when updating an existing PR**

The update path only changes title/body and adds labels. Therefore `draft` is ignored on reruns, and filtering the lookup by `--base` means changing the configured base can fail to find the existing PR for the deterministic branch and attempt to create another one. #854 explicitly requires updating the same PR and applying caller-supplied draft state. Find the open PR by head branch, retarget it to `BASE`, and transition it to draft/ready to match `DRAFT`.

dceoy (commented):
The implementation is well-structured and CI is green, but three issues should be addressed before merge: the actionlint artifact is no longer checksum-verified, the post-QA dirty check can miss repository changes, and existing PR draft/base state is not synchronized as required by #854.

.github/workflows/shell-project-ci.yml
dceoy: Fixed in 20c2ce7: the actionlint tarball is now downloaded to `RUNNER_TEMP` and verified against `actionlint_<version>_checksums.txt` with `sha256sum --check --strict` before extraction, mirroring the shfmt step.

.github/workflows/shell-project-ci.yml
dceoy: Fixed in 20c2ce7: the cleanliness check now runs `git -C "$GITHUB_WORKSPACE" status --porcelain --untracked-files=no`, covering the whole repo and both the index and worktree.

.github/actions/create-generated-update-pr/scripts/create-or-update-pr.sh
dceoy: Fixed in 20c2ce7: the update path now looks up the PR by head branch only, retargets it with `gh pr edit --base`, and toggles `gh pr ready`/`--undo` to match the draft input.

dceoy:
Applied the BASE injection-validation fix from this review in 20c2ce7. Deferring the force-push tree comparison (`origin/BRANCH` isn't guaranteed fresh right after `git fetch`, and the no-op case is already short-circuited by `stage-changes.sh`'s `changed=false`), the redundant `gh pr view` removal (now needed anyway for the draft-state check), and the six `: "${VAR:?}"` required-input tests as non-blocking follow-ups.

.github/workflows/shell-project-ci.yml
```diff
@@ -78,14 +55,22 @@ defaults:
...
+  ci:
     runs-on: ${{ inputs.runs-on }}
     timeout-minutes: ${{ inputs.timeout-minutes }}
     env:
       # renovate: datasource=github-releases depName=koalaman/shellcheck
       SHELLCHECK_VERSION: v0.10.0
       # renovate: datasource=github-releases depName=mvdan/sh
       SHFMT_VERSION: v3.12.0
+      # renovate: datasource=github-releases depName=rhysd/actionlint
+      ACTIONLINT_VERSION: v1.7.7
```
dceoy: **[HIGH] Enforce pinned native-tool versions, not only command availability** (confidence: high)

The workflow advertises centrally pinned versions, but the install steps return early whenever a command already exists. Both the default `ubuntu-slim` runner and the supported `ubuntu-24.04` runner already provide ShellCheck 0.9.0, so normal runs skip the configured `SHELLCHECK_VERSION=v0.10.0`; the pin and cache key are ineffective and checks depend on runner-image drift.

Always install the selected versions into a controlled directory, or compare each detected version with the requested version and install on mismatch. Add a test that the caller command observes the exact configured ShellCheck/shfmt/actionlint versions.

.github/actions/create-generated-update-pr/scripts/stage-changes.sh
```diff
@@ -0,0 +1,25 @@
...
+while IFS= read -r pathspec; do
+  [[ -n "${pathspec}" ]] && pathspecs+=("${pathspec}")
+done <<< "${PATHS}"
+
+if [[ ${#pathspecs[@]} -eq 0 ]]; then
+  echo "::error::paths input produced no pathspecs" >&2
+  exit 1
+fi
+
+git add -A -- "${pathspecs[@]}"
```
dceoy: **[MEDIUM] Treat an unmatched pathspec as a successful no-change run** (confidence: high)

`git add -A -- 'optional/generated/**'` exits with status 128 when the pathspec matches no tracked or untracked files. A scheduled generator that emits no optional output therefore fails before producing `changed=false`, contrary to the action's no-change acceptance criterion.

Check scoped status before staging, emit `changed=false` when it is empty, and only run `git add` when scoped changes exist. Add a Bats case where `PATHS` matches nothing.

.github/actions/create-generated-update-pr/scripts/create-or-update-pr.sh
```diff
@@ -0,0 +1,61 @@
...
+  fi
+  pr_number="${existing_pr_number}"
+else
+  create_args=(--title "${TITLE}" --body "${BODY:-}" --base "${BASE}" --head "${BRANCH}")
+  [[ "${draft}" == "true" ]] && create_args+=(--draft)
+  for label in "${labels[@]}"; do
+    create_args+=(--label "${label}")
+  done
+  gh pr create "${create_args[@]}"
+  pr_number="$(gh pr view "${BRANCH}" --json number --jq '.number')"
```
dceoy: **[LOW] Avoid resolving a new PR through an ambiguous positional argument** (confidence: high)

All-numeric branch names pass `git check-ref-format --branch`, but `gh pr view "123"` interprets `123` as pull request number rather than the head branch. For such a branch, the action can report an unrelated PR or fail after successfully creating the intended PR.

Capture the URL/number returned by creation, or query it explicitly with `gh pr list --head "${BRANCH}"` instead of `gh pr view "${BRANCH}"`.

dceoy (commented):
Reviewed head `20c2ce7`. CI is green and the previously reported issues are resolved, but two correctness issues remain in the default execution paths; one lower-severity branch-name edge case is also noted inline.

.github/workflows/shell-project-ci.yml
dceoy: Fixed in 1fcef7d: all three installs now check the pinned version of the binary at `~/.local/bin` directly and reinstall on mismatch, so a preinstalled runner binary no longer bypasses the pin (a cache hit still short-circuits since the cache key already encodes the exact versions). Added bats tests that stub the version check to assert the skip-vs-reinstall decision — not an end-to-end check of the real caller command's tool versions, which would need actual release downloads in CI. Note this does mean `enable-cache: false` callers now always download all three tools instead of short-circuiting on a preinstalled binary.

.github/actions/create-generated-update-pr/scripts/stage-changes.sh
dceoy: Fixed in 1fcef7d: stage-changes.sh now checks `git status --porcelain -- "${pathspecs[@]}"` first and reports changed=false without staging when nothing matches, avoiding git add's exit 128 on an unmatched pathspec. Added bats tests for both the unmatched-pathspec case and a new untracked file matching the pathspec.

.github/actions/create-generated-update-pr/scripts/create-or-update-pr.sh
dceoy: Fixed in 1fcef7d: the create path now resolves the new PR number via `gh pr list --head "${BRANCH}"` instead of `gh pr view "${BRANCH}"`, avoiding ambiguous resolution for all-numeric branch names. Added a regression test with BRANCH=123.

.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
```diff
@@ -0,0 +1,40 @@
...
+if [[ ${#pathspecs[@]} -eq 0 ]]; then
+  echo "::error::paths input produced no pathspecs" >&2
+  exit 1
+fi
+
+gh auth setup-git
+
+git checkout -B "${BRANCH}"
+git -c user.name="${GIT_USER_NAME}" -c user.email="${GIT_USER_EMAIL}" commit --only --message "${COMMIT_MESSAGE}" -- "${pathspecs[@]}"
+git push --force --set-upstream origin "${BRANCH}"
```
dceoy: **[P1] Refuse destructive branch targets before the force-push**

`BRANCH` is only syntax-validated. A caller typo such as `branch: main` passes this check, then this line force-pushes the generated commit to `origin/main` before `create-or-update-pr.sh` can fail because the head and base are identical. The same mechanism can overwrite any existing non-automation branch selected by mistake.

Pass `BASE` into this script and reject `BRANCH == BASE` before checkout/push. Preferably also protect the broader case by using a clearly managed branch namespace and `--force-with-lease` against a fetched remote tip rather than unconditional `--force`. Add a regression test proving that the base ref is unchanged when the inputs collide.

.github/actions/create-generated-update-pr/scripts/create-or-update-pr.sh
```diff
@@ -0,0 +1,61 @@
...
+labels=()
+if [[ -n "${LABELS:-}" ]]; then
+  while IFS= read -r label; do
+    [[ -n "${label}" ]] && labels+=("${label}")
+  done < <(tr ',' '\n' <<< "${LABELS}")
+fi
+
+draft="${DRAFT:-false}"
+
+existing_pr_number="$(gh pr list --head "${BRANCH}" --state open --json number --jq '.[0].number // empty')"
```
dceoy: **[P2] Restrict the PR lookup to the same-repository head**

`gh pr list --head "${BRANCH}"` filters only by branch name; the CLI does not support `<owner>:<branch>` for this flag. An open PR from a fork can therefore use the same deterministic branch name and become `.[0]`, causing this action to edit and report the wrong PR instead of the branch it just pushed to `origin`.

Request `isCrossRepository`/`headRepositoryOwner` in the JSON and select the same-repository PR explicitly (and apply the same filter after `gh pr create`). Add a collision test with a fork PR using the same head branch name.

dceoy (commented):
Reviewed head `1fcef7d`. CI is green and all previously raised threads are resolved, but two safety/correctness issues remain in the generated-update action. The force-push path can overwrite a caller-selected existing branch, including the PR base, and the PR lookup can select a fork PR that happens to use the same branch name.

.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
dceoy: Fixed in 6db6d19: commit-and-push.sh now requires `BASE` and refuses to force-push when `BRANCH` equals `BASE` (regression test added). Force-pushing the caller-selected `branch` itself remains intentional per the action's "create or reset" contract, so `--force-with-lease`/namespace restriction is not added here.

.github/actions/create-generated-update-pr/scripts/create-or-update-pr.sh
dceoy: Fixed in 6db6d19: the PR lookup now filters to `isCrossRepository == false` in both the update and post-create paths, so a fork PR reusing the same head branch name is no longer selected (regression test added).

dceoy (commented):
Three blocking findings remain on the current head. GitHub does not allow requesting changes on one's own pull request, so these are submitted as review comments rather than a REQUEST_CHANGES review.

.github/workflows/shell-project-ci.yml
```diff
@@ -3,53 +3,20 @@ name: Run shell project CI
...
-        default: true
-      bats-test-paths:
+      python-version:
         required: false
         type: string
-        description: Newline-separated Bats test pathspecs passed to git ls-files
-        default: |
-          *.bats
+        description: Python version used to install zizmor, yamllint, and Checkov
+        default: 3.x
```
dceoy: **[P1] Pin Python within Checkov's supported range**

`3.x` currently resolves to Python 3.14, while Checkov 3.3.1 documents support only through Python 3.13 (and lists classifiers only through 3.12). The minimal/default workflow can therefore install and run Checkov on an unsupported interpreter, despite the caller not opting into a newer Python. Use a supported fixed minor such as `3.12`, or a bounded setup-python range such as `>=3.10 <3.14`, and test the default workflow end to end.

.github/workflows/shell-project-ci.yml
```diff
@@ -78,14 +55,22 @@ defaults:
...
       # renovate: datasource=github-releases depName=mvdan/sh
       SHFMT_VERSION: v3.12.0
+      # renovate: datasource=github-releases depName=rhysd/actionlint
+      ACTIONLINT_VERSION: v1.7.7
+      # renovate: datasource=pypi depName=zizmor versioning=pep440
+      ZIZMOR_VERSION: "1.26.1"
+      # renovate: datasource=pypi depName=yamllint versioning=pep440
+      YAMLLINT_VERSION: "1.38.0"
+      # renovate: datasource=pypi depName=checkov versioning=pep440
+      CHECKOV_VERSION: "3.3.1"
```
dceoy: **[P2] Do not downgrade the initial consumer's scanners**

The stated initial consumer, `oracle-pr-sentry`, currently pins zizmor 1.28.0, Checkov 3.3.8, and actionlint 1.7.12, but this workflow centralizes older 1.26.1, 3.3.1, and 1.7.7 releases. Migrating it would silently downgrade security and workflow analysis, conflicting with the acceptance criterion that the consumer migrate without losing current CI behavior. Align these pins with at least the consumer's current versions before replacing its local installation.

.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
```diff
@@ -0,0 +1,46 @@
...
+done <<< "${PATHS}"
+
+if [[ ${#pathspecs[@]} -eq 0 ]]; then
+  echo "::error::paths input produced no pathspecs" >&2
+  exit 1
+fi
+
+gh auth setup-git
+
+git checkout -B "${BRANCH}"
```
dceoy, line 53: **[P2] Ensure the update branch is based on the requested base**

`BASE` is only used for the equality guard; `git checkout -B "$BRANCH"` resets the branch from whatever commit the caller happened to check out. With the documented default checkout, setting `base: develop` still creates the update branch from the default branch and then opens/retargets a PR against `develop`, potentially including unrelated commits. Either require and validate that `HEAD` is the requested base before mutating the remote, or redesign the flow so generation starts from `origin/$BASE`; document that invariant and add a regression test.

.github/workflows/shell-project-ci.yml
dceoy: Fixed in be7c8e2: `python-version` default is now `3.12`, which Checkov 3.3.x (and zizmor/yamllint) support per PyPI classifiers/`requires_python`; `3.x` was resolving to unsupported 3.14.

.github/workflows/shell-project-ci.yml
dceoy: Fixed in be7c8e2: bumped to zizmor 1.28.0, Checkov 3.3.8, and actionlint v1.7.12 to match oracle-pr-sentry's current pins; verified via PyPI metadata that both zizmor and Checkov support the new python-version 3.12 default.

.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
dceoy: Fixed in be7c8e2: commit-and-push.sh now compares HEAD against base via the GitHub compare API and only proceeds when the status is `identical` or `behind` (rejecting `ahead`/`diverged`, e.g. your `base: develop` with main checked out example); `behind` is accepted so base advancing between checkout and this step (a benign race for scheduled generators) doesn't fail the run. Starting generation from `origin/$BASE` isn't possible here since this action runs after the caller's own generation step, so validation was the only available option. Compare is scoped to $GITHUB_REPOSITORY while the push targets `origin`; these match under the documented checkout usage. Regression tests added for both the diverged and behind cases, and the invariant is documented in README.md.tmpl.

.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
```diff
@@ -0,0 +1,57 @@
...
+: "${BASE:?BASE is required}"
+: "${COMMIT_MESSAGE:?COMMIT_MESSAGE is required}"
+: "${GIT_USER_NAME:?GIT_USER_NAME is required}"
+: "${GIT_USER_EMAIL:?GIT_USER_EMAIL is required}"
+
+if [[ -z "${GH_TOKEN:-}" ]]; then
+  echo "::error::GH_TOKEN must be set in the step environment" >&2
+  exit 1
+fi
+
```
dceoy, line 14: **[P1] Reject Git checkout shorthand before validating the branch target**

`git check-ref-format --branch` expands `@{-n}` using the checkout reflog. As a result, `BRANCH='@{-1}'` can pass this validation while the script keeps the unexpanded string; `git checkout -B "${BRANCH}"` then expands it again and resets the previously checked-out branch. This also bypasses the literal `BRANCH == BASE` guard, so a caller that previously checked out `main` can force-push `main`. Validate `refs/heads/${BRANCH}` without `--branch` (or explicitly reject `@{-n}`) and add a regression test proving the base ref is not modified.

.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
```diff
@@ -0,0 +1,57 @@
...
+  identical | behind) ;;
+  *)
+    echo "::error::the checked-out commit is not based on base (${BASE}); checkout base before running this action" >&2
+    exit 1
+    ;;
+esac
+
+git checkout -B "${BRANCH}"
+git -c user.name="${GIT_USER_NAME}" -c user.email="${GIT_USER_EMAIL}" commit --only --message "${COMMIT_MESSAGE}" -- "${pathspecs[@]}"
+git push --force --set-upstream origin "${BRANCH}"
```
dceoy: **[P2] Avoid force-pushing an identical generated tree on every rerun**

Scheduled reruns start from the base branch, so scoped generated files can differ from the base even when the existing update branch already contains exactly the same generated result. This creates a new commit timestamp and unconditionally force-pushes a different SHA with the same tree, firing `synchronize` and rerunning all PR CI each time. Fetch the managed branch and skip the push when its tree already matches the newly created tree; in that case report the remote commit SHA. Add a two-run regression test with unchanged generated output.

dceoy (commented):
Two correctness issues remain in the generated-update action: branch shorthand can bypass the destructive-target guard, and unchanged generated output still rewrites the remote branch on every scheduled run.

.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
dceoy: Fixed in 0d4b326: branch validation now uses `git check-ref-format refs/heads/${BRANCH}` (no `--branch`), so reflog shorthand like `@{-1}` is rejected instead of expanded past the literal `BRANCH == BASE` guard; regression test added.

.github/actions/create-generated-update-pr/scripts/commit-and-push.sh
dceoy: Fixed in 0d4b326: the script now fetches the existing branch and compares tree hashes before pushing, skipping the force-push and reporting the existing remote SHA when the generated tree is unchanged; regression test added. (This reverses the earlier deferral on the same finding from the OpenCode review comment, whose rationale didn't hold up.)

dceoy:
Reversing the earlier deferral here: fixed in 0d4b326, which fetches the existing branch and compares tree hashes before force-pushing, skipping the push and reporting the existing remote SHA when the generated tree is unchanged. A second, independent review flagged the same issue and the original deferral rationale (that `stage-changes.sh`'s `changed=false` already short-circuits this) doesn't hold, since that check compares against `base`, not the update branch's previous tip.

dceoy:
Addressed the three OpenCode findings not already covered by the earlier deferral (BASE validation and the force-push tree comparison were already fixed): added regression tests for `--draft` on the create path, blank-only `PATHS`, and a non-existent `search-path` (b37cd6c). The `gh pr view` removal and the six required-input tests remain deferred as non-blocking per the earlier comment.

.github/actions/create-generated-update-pr/action.yml
```diff
@@ -0,0 +1,93 @@
...
+        BRANCH: ${{ inputs.branch }}
+        BASE: ${{ inputs.base }}
+        COMMIT_MESSAGE: ${{ inputs.commit-message }}
+        GIT_USER_NAME: ${{ inputs.git-user-name }}
+        GIT_USER_EMAIL: ${{ inputs.git-user-email }}
+        PATHS: ${{ inputs.paths }}
+      run: ${{ github.action_path }}/scripts/commit-and-push.sh
+    - name: Create or update the pull request
+      id: pr
+      if: steps.stage.outputs.changed == 'true'
```
dceoy, line 84: **[P2] Close an obsolete update PR when the scoped diff disappears**

Both commit/push and PR reconciliation are skipped when `changed == 'false'`. If a previous run opened an update PR and a later generator run produces content identical to `base`, this action exits successfully but leaves the old PR open with its now-obsolete diff. That stale PR can then be merged and reintroduce an update the latest generation explicitly no longer proposes. On the no-change path, look up the same-repository open PR for `branch` and close it (and optionally delete the managed branch) without pushing; add a regression test covering changed → open PR → no-change.

dceoy (commented):
One blocking lifecycle issue remains in the generated-update action. The current no-change path skips all PR reconciliation, so an obsolete open update PR can survive after the generator returns to the base tree.

.github/actions/create-generated-update-pr/action.yml
dceoy: Fixed in 5f2627d: added a `close-obsolete-pr.sh` step guarded by `changed == 'false'` that closes the same-repo open PR for `branch` (fork PRs excluded, same as the create/update lookup). Branch deletion is intentionally left out — keeping the branch around means a wrongly-closed PR stays recoverable. Added a bats regression test covering changed -> open PR -> no-change.

README.md.tmpl
```diff
@@ -93,27 +93,47 @@ jobs:
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+      - run: ./scripts/generate-docs.sh
+      - uses: dceoy/gh-actions-for-devops/.github/actions/create-generated-update-pr@main
```
dceoy: **[P1] Reference the current action repository name**

This newly added example still calls `dceoy/gh-actions-for-devops`, while the repository is now `dceoy/gha-for-devops`. GitHub Actions intentionally does not follow repository redirects for actions or reusable workflows, so callers copying this snippet receive `repository not found`. Update both the generated-action and shell-workflow examples to `dceoy/gha-for-devops`, then regenerate `README.md`.

dceoy (commented):
The implementation and CI are otherwise in good shape, but the newly documented caller snippets still use the repository's former name. GitHub Actions does not follow repository redirects for actions or reusable workflows, so these examples are not usable after the rename.

README.md.tmpl
dceoy: Fixed in 3a8cda0: both the generated-action and shell-workflow examples in README.md.tmpl now reference `dceoy/gha-for-devops`, and README.md is regenerated.

dceoy (commented):
Reviewed head `3a8cda0`. No blocking findings remain. The previously reported scope-safety, branch validation, PR lifecycle, pinned-tool, and documentation issues are resolved, and the current CI run is green. The only residual risk is non-blocking: `shell-project-ci.yml` is covered through extracted-step Bats tests rather than an end-to-end reusable-workflow smoke job, so the first downstream migration should validate the complete installation and caller-command path on the selected runner.

Issue #865, opened 2026-07-31
Add a trusted reusable repository PR security gate


dceoy:
## Resolution — not planned

This proposal is obsolete for `gha-for-devops`.

The repository did implement this direction in #866–#873, but #913 subsequently removed the repository-security workflow, composite action, scanner implementation, configuration, fixtures, tests, and documentation after deciding that a centrally owned repository PR security scanner is outside this repository's intended reusable-DevOps-workflow scope.

Reintroducing the same subsystem here would reverse that explicit scope decision. Any future central repository-security gate should be tracked in the repository that owns that security policy rather than reopening this implementation in `gha-for-devops` without a new scope decision.

---

## Historical proposal

`dceoy/segh` owned a central pull-request security workflow in addition to its organization governance audit CLI. This issue proposed moving that generic CI/security infrastructure into `gha-for-devops` and providing one trusted, read-only, repository-agnostic PR gate with a stable required check and retained evidence.

The historical implementation requirements included Zizmor, actionlint with embedded ShellCheck, standalone ShellCheck, Checkov, Trivy vulnerability/secret scans, read-only permissions, pinned scanner policy, fork/Dependabot support, merge-group handling, complete evidence retention, and fixture-backed failure-mode coverage.

That implementation landed and was later intentionally removed by #913, so the original acceptance criteria are no longer active.

PR #866, opened 2026-07-31
Add trusted repository security scan workflow

dceoy:
## Summary

- add a reusable, workflow_call-only `Repository security / scan` workflow, composite action, scanner script, configuration, and Aqua pins
- run centrally pinned zizmor, actionlint with embedded ShellCheck, standalone ShellCheck, Checkov, and separate Trivy vulnerability and secret gates against an isolated target checkout
- remove tracked symlinks before scanning, retain bounded summaries plus complete JSON/text/log/status evidence, and fail closed through one aggregate result
- document reusable-workflow usage and the current bootstrap caveat for organization-ruleset usage
- add fixture-backed coverage for clean, failing, empty, symlink, simultaneous-failure, merge-group, and fork/read-only scenarios

## Why

The repository did not have one trusted, repository-agnostic PR security gate suitable for organization rulesets. Existing workflows covered individual tools but exposed different permissions, configuration, and reporting contracts. This change moves the generic gate owned by `dceoy/segh` into the reusable-workflow repository while keeping scanner policy in the trusted workflow revision.

## Staged rollout (bootstrap)

This PR keeps the `scan` job's trust boundary as designed: for a direct `pull_request` trigger on this repository's own workflow, the job checks out its **base** revision as the trusted scanner implementation, never the untrusted PR head, and there is no fallback from base to head. Because the trusted implementation (`.github/actions/repository-security-scan/action.yml`) does not exist on any commit before this PR, a direct `pull_request`/`merge_group` trigger on this same workflow would fail before any scanner runs (the base revision has nothing to check out).

To avoid that bootstrap deadlock without weakening the boundary:

1. **This PR** adds the workflow, action, scanner script, configuration, tests, Aqua pins, and docs, but enables **only `workflow_call`** — no direct `pull_request`/`merge_group` triggers yet.
2. Once merged, the implementation exists at a reviewed commit on `main` and can already be consumed via `workflow_call` at a pinned SHA.
3. **Follow-up: #867** re-enables direct `pull_request`/`merge_group` triggers on this workflow once a trusted revision exists on `main`, restoring direct organization-ruleset usage.
4. **#865 stays open** until #867 merges — the organization-ruleset required-check use case it asked for isn't available until direct triggers are re-enabled.

## Impact

Consumers can call `.github/workflows/repository-security-scan.yml` via `workflow_call` at a reviewed commit SHA today. Direct organization-ruleset selection (`pull_request`/`merge_group`) lands in #867. The stable check name is `Repository security / scan`; it needs only `contents: read`, no secrets, no GHAS features, and supports fork, Dependabot, and merge-queue revisions.

## Validation

- `bats --print-output-on-failure .github/scripts/tests/*.bats` (88 tests; 2 pre-existing failures are macOS BSD `sed` incompatibilities in local fixture setup, unrelated to this change and not present on the `ubuntu-24.04` CI runner)
- `actionlint` and `zizmor --offline --no-config` against the updated workflow (clean)
- `.github/scripts/update-readme.sh` (README regenerated cleanly)
- real-scanner dry run against an index-built copy of this repository (all six scanner statuses and preflight passed)
- security pattern scan and manual CI/CD trust-boundary review

Refs #865. Follow-up: #867.

### DIFF
--- a/.github/actions/repository-security-scan/action.yml
+++ b/.github/actions/repository-security-scan/action.yml
@@ -1,0 +1,59 @@
+---
+name: Trusted repository security scanner pipeline
+description: Run centrally configured security gates against an untrusted repository checkout
+runs:
+  using: composite
+  steps:
+    - name: Install checksum-verified scanners
+      uses: aquaproj/aqua-installer@96a9bc20066c5bf5e275b41019cfc165b25f4e2e  # v4.0.5
+      with:
+        aqua_version: v2.62.1
+        working_directory: _trusted
+    - name: Prepare evidence directory
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" prepare'
+    - name: Remove tracked symlinks from the target checkout
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" preflight'
+    - name: Run zizmor gate
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" zizmor'
+    - name: Run actionlint and embedded ShellCheck gate
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" actionlint'
+    - name: Run standalone ShellCheck gate
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" shellcheck'
+    - name: Run Checkov infrastructure-as-code gate
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" checkov'
+    - name: Run Trivy vulnerability gate
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" trivy-vulnerability'
+    - name: Run Trivy secret gate
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" trivy-secret'
+    - name: Publish bounded scanner summary
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" summary'
+    - name: Retain complete scanner evidence
+      if: always()
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
+      with:
+        name: repository-security-reports
+        path: security-results
+        if-no-files-found: error
+        retention-days: 14
+    - name: Enforce aggregate scanner result
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" enforce'
--- a/.github/actions/repository-security-scan/action.yml
+++ b/.github/actions/repository-security-scan/action.yml
@@ -13,7 +13,7 @@ runs:
       if: always()
       shell: bash
       run: '"$GITHUB_ACTION_PATH/scan.sh" prepare'
-    - name: Remove tracked symlinks from the target checkout
+    - name: Validate the target checkout
       if: always()
       shell: bash
       run: '"$GITHUB_ACTION_PATH/scan.sh" preflight'
--- a/.github/actions/repository-security-scan/action.yml
+++ b/.github/actions/repository-security-scan/action.yml
@@ -49,7 +49,7 @@ runs:
       if: always()
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
       with:
-        name: repository-security-reports
+        name: repository-security-reports-${{ github.run_attempt }}
         path: security-results
         if-no-files-found: error
         retention-days: 14
--- a/.github/actions/repository-security-scan/scan.sh
+++ b/.github/actions/repository-security-scan/scan.sh
@@ -210,6 +210,48 @@ reject_script_shellcheck_directives() {
   return 1
 }
 
+extract_composite_action_shell_blocks() {
+  local output_name=$1
+  shift
+  local -n composite_output_ref=${output_name}
+  local action_file block_json block_index decode_status extracted_file
+  local blocks_jsonl="${results_dir}/shellcheck-composite-blocks.jsonl"
+  local extraction_dir="${results_dir}/shellcheck-composite"
+  local shell_pattern='(^|[[:space:]/])(sh|ash|dash|bash|ksh93|ksh88|ksh|oksh|bats)([[:space:]]|$)'
+
+  composite_output_ref=()
+  for action_file in "$@"; do
+    if ! yq eval --output-format=json --unwrapScalar=false \
+      "select(.runs.using == \"composite\") | .runs.steps[]? | select(has(\"run\")) | select(.shell | test(\"${shell_pattern}\")) | .run" \
+      "${action_file}" > "${blocks_jsonl}" 2>> "${results_dir}/shellcheck.log"; then
+      rm -f -- "${blocks_jsonl}"
+      rm -rf -- "${extraction_dir}"
+      return 1
+    fi
+
+    block_index=0
+    decode_status=0
+    while IFS= read -r block_json; do
+      extracted_file="${extraction_dir}/${action_file}/step-${block_index}.sh"
+      mkdir -p "$(dirname "${extracted_file}")"
+      if ! printf '%s\n' "${block_json}" \
+        | yq eval --input-format=json --unwrapScalar '.' - \
+          > "${extracted_file}" 2>> "${results_dir}/shellcheck.log"; then
+        decode_status=1
+        break
+      fi
+      composite_output_ref+=("${extracted_file}")
+      ((block_index += 1))
+    done < "${blocks_jsonl}"
+    if ((decode_status != 0)); then
+      rm -f -- "${blocks_jsonl}"
+      rm -rf -- "${extraction_dir}"
+      return 1
+    fi
+  done
+  rm -f -- "${blocks_jsonl}"
+}
+
 run_zizmor() {
   local scanner_status_value render_status_value
   local -a workflow_files=()
@@ -318,6 +360,9 @@ run_actionlint() {
 run_shellcheck() {
   local scanner_status_value render_status_value entry mode file first_line
   local shell_shebang='^#![[:space:]]*(/usr/bin/env[[:space:]]+([^[:space:]]+[[:space:]]+)*)?(/[^[:space:]]*/)?(busybox[[:space:]]+)?(sh|ash|dash|bash|ksh93|ksh88|ksh|oksh|bats)([[:space:]]|$)'
+  local composite_shell_dir="${results_dir}/shellcheck-composite"
+  local -a action_files=()
+  local -a composite_shell_files=()
   local -a script_files=()
   local -a unreadable_files=()
 
@@ -328,6 +373,9 @@ run_shellcheck() {
     file=${entry#*$'\t'}
     if [[ "${mode}" != '100644' && "${mode}" != '100755' ]]; then
       continue
+    fi
+    if [[ "${file}" =~ (^|.*/)action\.ya?ml$ ]]; then
+      action_files+=("${file}")
     fi
     if [[ "${file}" == *.sh || "${file}" == *.bash || "${file}" == *.bats ]]; then
       script_files+=("${file}")
@@ -355,11 +403,23 @@ run_shellcheck() {
     printf '1\n' > "${results_dir}/shellcheck.status"
     return 0
   fi
+
+  : > "${results_dir}/shellcheck.log"
+  if ! extract_composite_action_shell_blocks composite_shell_files "${action_files[@]}"; then
+    printf '{"scanner":"shellcheck","result":"not-run","reason":"unable to inspect composite action shell blocks"}\n' \
+      > "${results_dir}/shellcheck.json"
+    printf 'Failed: unable to inspect composite action shell blocks.\n' \
+      > "${results_dir}/shellcheck.txt"
+    printf '1\n' > "${results_dir}/shellcheck.status"
+    return 0
+  fi
+  script_files+=("${composite_shell_files[@]}")
   if ((${#script_files[@]} == 0)); then
-    write_skip_evidence shellcheck 'no tracked standalone shell scripts were found.'
+    write_skip_evidence shellcheck 'no tracked standalone shell scripts or composite action shell blocks were found.'
     return 0
   fi
   if reject_script_shellcheck_directives shellcheck "${script_files[@]}"; then
+    rm -rf -- "${composite_shell_dir}"
     return 0
   fi
 
@@ -384,6 +444,7 @@ run_shellcheck() {
     > "${results_dir}/shellcheck.txt" \
     2>> "${results_dir}/shellcheck.log"
   render_status_value=$?
+  rm -rf -- "${composite_shell_dir}"
   if ((scanner_status_value == 0 && render_status_value != 0)); then
     scanner_status_value=${render_status_value}
   fi
--- a/.github/actions/repository-security-scan/scan.sh
+++ b/.github/actions/repository-security-scan/scan.sh
@@ -130,6 +130,86 @@ collect_regular_files() {
   done < "${results_dir}/tracked-files.index"
 }
 
+write_shellcheck_directive_failure() {
+  local scanner_name="$1"
+  local directive_report="$2"
+
+  printf '{"scanner":"%s","result":"policy-violation","reason":"target-owned ShellCheck directives"}\n' \
+    "${scanner_name}" > "${results_dir}/${scanner_name}.json"
+  {
+    printf 'Failed: target-owned ShellCheck directives are not allowed; exclusions belong in the trusted policy.\n'
+    sed 's/^/  /' "${directive_report}"
+  } > "${results_dir}/${scanner_name}.txt"
+  printf '1\n' > "${results_dir}/${scanner_name}.status"
+}
+
+reject_workflow_shellcheck_directives() {
+  local scanner_name="$1"
+  shift
+  local file match
+  local directive_pattern='^[[:space:]]*#[[:space:]]*shellcheck[[:space:]]+[[:alpha:]-]+='
+  local directive_report="${results_dir}/${scanner_name}.shellcheck-directives"
+  local run_blocks="${results_dir}/${scanner_name}.run-blocks"
+  local matches="${results_dir}/${scanner_name}.shellcheck-matches"
+
+  : > "${directive_report}"
+  : > "${results_dir}/${scanner_name}.log"
+  for file in "$@"; do
+    if ! yq eval '.jobs[]?.steps[]? | select(has("run")) | .run' "${file}" \
+      > "${run_blocks}" 2>> "${results_dir}/${scanner_name}.log"; then
+      printf '{"scanner":"%s","result":"not-run","reason":"unable to inspect workflow shell blocks"}\n' \
+        "${scanner_name}" > "${results_dir}/${scanner_name}.json"
+      printf 'Failed: unable to inspect workflow shell blocks for target-owned ShellCheck directives.\n' \
+        > "${results_dir}/${scanner_name}.txt"
+      printf '1\n' > "${results_dir}/${scanner_name}.status"
+      rm -f -- "${directive_report}" "${run_blocks}" "${matches}"
+      return 0
+    fi
+    if LC_ALL=C grep -nE "${directive_pattern}" "${run_blocks}" > "${matches}"; then
+      while IFS= read -r match; do
+        printf '%s:%s\n' "${file}" "${match}" >> "${directive_report}"
+      done < "${matches}"
+    fi
+  done
+  rm -f -- "${run_blocks}" "${matches}"
+  if [[ -s "${directive_report}" ]]; then
+    write_shellcheck_directive_failure "${scanner_name}" "${directive_report}"
+    rm -f -- "${directive_report}"
+    return 0
+  fi
+  rm -f -- "${directive_report}" "${results_dir}/${scanner_name}.log"
+  return 1
+}
+
+reject_script_shellcheck_directives() {
+  local scanner_name="$1"
+  shift
+  local grep_status
+  local directive_pattern='^[[:space:]]*#[[:space:]]*shellcheck[[:space:]]+[[:alpha:]-]+='
+  local directive_report="${results_dir}/${scanner_name}.shellcheck-directives"
+
+  : > "${results_dir}/${scanner_name}.log"
+  LC_ALL=C grep -nHE "${directive_pattern}" -- "$@" \
+    > "${directive_report}" 2>> "${results_dir}/${scanner_name}.log"
+  grep_status=$?
+  if ((grep_status == 0)); then
+    write_shellcheck_directive_failure "${scanner_name}" "${directive_report}"
+    rm -f -- "${directive_report}"
+    return 0
+  fi
+  if ((grep_status != 1)); then
+    printf '{"scanner":"%s","result":"not-run","reason":"unable to inspect shell scripts"}\n' \
+      "${scanner_name}" > "${results_dir}/${scanner_name}.json"
+    printf 'Failed: unable to inspect shell scripts for target-owned ShellCheck directives.\n' \
+      > "${results_dir}/${scanner_name}.txt"
+    printf '1\n' > "${results_dir}/${scanner_name}.status"
+    rm -f -- "${directive_report}"
+    return 0
+  fi
+  rm -f -- "${directive_report}" "${results_dir}/${scanner_name}.log"
+  return 1
+}
+
 run_zizmor() {
   local scanner_status_value render_status_value
   local -a workflow_files=()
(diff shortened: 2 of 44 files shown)


dceoy (commented):
The trust-boundary design is generally sound, but the workflow cannot bootstrap in the PR that introduces it: the trusted base revision does not contain the newly added composite action, so the new required check fails before any scanner runs. Please split the rollout or add an explicit safe bootstrap path, then verify the actual `Repository security / scan` job succeeds.

.github/workflows/repository-security-scan.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - name: Check out trusted scanner implementation
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          repository: ${{ job.workflow_repository }}
+          # A pull request against this source repository must use its base
+          # revision as the scanner. Reusable and ruleset runs use the exact
+          # repository and SHA of the workflow that defines this job, not the
+          # caller workflow identity from the github context.
+          ref: ${{ job.workflow_repository == github.repository && github.event_name == 'pull_request' && github.event.pull_request.base.sha || job.workflow_sha }}
```
dceoy: **Blocking:** for this introducing PR, this expression resolves to the base SHA (`32d130e…`). That revision does not contain `./.github/actions/repository-security-scan`, so the later local-action step fails immediately with `Can't find 'action.yml'...`; the current `Repository security` run is red before any scanner executes. Bootstrap this in two stages (land the action/config first, then enable the `pull_request`/`merge_group` workflow) or implement an explicit safe first-install path. Keep subsequent scanner changes pinned to the base revision so PR-controlled code cannot replace the trusted implementation.

dceoy: Fixed in 3c16a31: only `workflow_call` is enabled now (pull_request/merge_group deferred to #867 once the trusted revision is on main), with a boundary test asserting it; CI is green.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,432 @@
...
+  collect_regular_files workflow_files github-actions
+  if ((${#workflow_files[@]} == 0)); then
+    write_skip_evidence zizmor 'no tracked GitHub Actions workflows or actions were found.'
+    return 0
+  fi
+
+  : > "${results_dir}/zizmor.log"
+  zizmor \
+    --offline \
+    --no-config \
```
dceoy, line 344: **P1: Disable inline zizmor suppressions**

`--no-config` prevents loading a target-owned `zizmor.yml`, but it still honors source comments such as `# zizmor: ignore[template-injection]`. A pull-request author can therefore suppress a security finding in the untrusted workflow/action and make this centrally enforced gate pass. zizmor v1.28 supports `--no-ignores`; add it to both the JSON and plain invocations, and cover the bypass with a fixture test.

dceoy (commented):
One additional policy-bypass issue remains in the trusted scanner path.

.github/actions/repository-security-scan/scan.sh
dceoy: Fixed in 200a2086cc: added `--no-ignores` to both zizmor invocations so a target-owned `# zizmor: ignore[...]` comment can no longer suppress a finding.

.github/security/repository-security/checkov.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - openapi
+  - serverless
+  - terraform
+  - terraform_json
+  - terraform_plan
+  - yaml
+
+# Add centrally accepted Checkov IDs only after review. Target-repository
+# configuration is never loaded.
+skip-check: []
```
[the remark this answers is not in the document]
dceoy: checkov has no `--skip-suppressions` flag (checked 3.3.x source and upstream docs/issues); fixed in 200a2086cc by failing the gate when `checkov.json` reports any `results.skipped_checks` entries, which can only originate from an inline `#checkov:skip=` comment since the trusted config's `skip-check` list is always empty — mirroring zizmor's `--no-ignores`.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,432 @@
...
+      continue
+    fi
+    case "${selection}" in
+      workflows)
+        [[ "${file}" =~ ^\.github/workflows/[^/]+\.ya?ml$ ]] && output_files+=("${file}")
+        ;;
+      github-actions)
+        if [[ "${file}" =~ ^\.github/workflows/[^/]+\.ya?ml$ ]] \
+          || [[ "${file}" =~ ^\.github/actions/(.*/)?action\.ya?ml$ ]]; then
+          output_files+=("${file}")
```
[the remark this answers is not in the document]
dceoy: Fixed in 200a2086cc: `collect_regular_files` now matches any tracked `action.yml`/`action.yaml`, not just ones under `.github/actions/`.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,432 @@
...
+    if [[ "${mode}" != '100644' && "${mode}" != '100755' ]]; then
+      continue
+    fi
+    case "${selection}" in
+      workflows)
+        [[ "${file}" =~ ^\.github/workflows/[^/]+\.ya?ml$ ]] && output_files+=("${file}")
+        ;;
+      github-actions)
+        if [[ "${file}" =~ ^\.github/workflows/[^/]+\.ya?ml$ ]] \
+          || [[ "${file}" =~ ^\.github/actions/(.*/)?action\.ya?ml$ ]]; then
```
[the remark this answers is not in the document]
dceoy: Same fix as the other local-action-path finding on this PR: `collect_regular_files` now matches any tracked `action.yml`/`action.yaml` regardless of parent directory (200a2086cc).

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,432 @@
...
+    mode=${entry%% *}
+    file=${entry#*$'\t'}
+    if [[ "${mode}" != '100644' && "${mode}" != '100755' ]]; then
+      continue
+    fi
+    if [[ "${file}" == *.sh || "${file}" == *.bash || "${file}" == *.bats ]]; then
+      script_files+=("${file}")
+      continue
+    fi
+    IFS= read -r -n 256 first_line < "${file}" || true
```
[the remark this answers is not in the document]
dceoy: Fixed in 200a2086cc: unreadable files now fail the shellcheck gate closed with an `::error::` annotation instead of being silently dropped from `script_files`.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,432 @@
...
+  if ((scanner_status_value == 0 && render_status_value != 0)); then
+    scanner_status_value=${render_status_value}
+  fi
+  ensure_text_evidence actionlint "${scanner_status_value}"
+  printf '%s\n' "${scanner_status_value}" > "${results_dir}/actionlint.status"
+}
+
+run_shellcheck() {
+  local scanner_status_value render_status_value entry mode file first_line
+  local shell_shebang='^#![[:space:]]*(/usr/bin/env([[:space:]]+-S)?[[:space:]]+)?(/[^[:space:]]*/)?(busybox[[:space:]]+)?(sh|ash|dash|bash|ksh93|ksh88|ksh|oksh|bats)([[:space:]]|$)'
```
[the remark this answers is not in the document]
dceoy: Fixed in 200a2086cc: the shebang regex now accepts any sequence of `env` options (e.g. `env -i`), not just `-S`.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,432 @@
...
+    actionlint \
+    --no-color \
+    --config-file "${trusted_dir}/.github/security/repository-security/actionlint.yaml" \
+    --shellcheck shellcheck \
+    --format '{{json .}}' \
+    "${workflow_files[@]}" \
+    > "${json_lines}" \
+    2>> "${results_dir}/actionlint.log"
+  scanner_status_value=$?
+  {
```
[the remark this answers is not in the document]
dceoy: Fixed in 200a2086cc: `actionlint.json` is now built with `yq eval-all` instead of manual bracket/comma joining, so blank lines in the JSON Lines stream can no longer produce invalid JSON.

.github/security/repository-security/trivy-secret.yaml
```diff
@@ -0,0 +1,4 @@
+# Use Trivy's complete built-in secret rule set without target-controlled
+# custom rules or allow rules.
+rules: []
+allow-rules: []
```
dceoy, line 4: **P1: Disable or explicitly accept Trivy’s built-in allow rules**

`allow-rules: []` only means no *custom* allow rules; it does not disable Trivy’s built-in allow rules. Trivy ignores Markdown secrets by default, so a PR can place a real credential in `README.md` and this supposedly repository-wide “fail on any secret” gate can pass. The comment above this line also incorrectly claims there are no allow rules.

Configure `disable-allow-rules` for the built-in exemptions that this policy must not honor (and review the default `skip-patterns` similarly), then add a fixture containing a detectable secret in Markdown. If those defaults are intentionally accepted, document that reduced coverage instead of claiming complete secret scanning.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,474 @@
...
+  if ((${#script_files[@]} == 0)); then
+    write_skip_evidence shellcheck 'no tracked standalone shell scripts were found.'
+    return 0
+  fi
+
+  : > "${results_dir}/shellcheck.log"
+  shellcheck \
+    --color=never \
+    --format=json1 \
+    --rcfile "${trusted_dir}/.github/security/repository-security/shellcheckrc" \
```
dceoy, line 514: **P1: Reject target-owned ShellCheck suppression directives**

A trusted `--rcfile` prevents a target `.shellcheckrc` from changing policy, but it does not disable source directives such as `# shellcheck disable=SC2086` or `# shellcheck disable=all`. A PR author can therefore suppress standalone ShellCheck findings—and directives inside workflow `run:` blocks can also affect actionlint’s embedded ShellCheck—despite the requirement that exclusions remain in the trusted revision.

Fail closed on target-owned ShellCheck directives or scan a directive-neutralized copy, and add fixtures proving both standalone scripts and workflow shell blocks cannot suppress findings. Otherwise the documentation should explicitly state that source-level suppressions are part of the accepted target-controlled policy.

dceoy (commented):
Two target-controlled policy-bypass paths remain in the trusted scanner. The current CI and Bats suite are green, and the previously reported bootstrap, zizmor, Checkov, local-action, shebang, unreadable-file, and JSONL issues are resolved, but these two cases need explicit policy handling and fixture coverage before merge.

.github/security/repository-security/trivy-secret.yaml
dceoy: Fixed in 5953686: the trusted Trivy policy now disables all 13 built-in allow rules and default skip patterns, with a Markdown-secret regression fixture.

.github/actions/repository-security-scan/scan.sh
dceoy: Fixed in 5953686: both workflow run blocks and standalone shell inputs now fail closed on target-owned ShellCheck directives, with regression fixtures for each path.

.github/actions/repository-security-scan/action.yml
```diff
@@ -0,0 +1,59 @@
...
+      run: '"$GITHUB_ACTION_PATH/scan.sh" trivy-secret'
+    - name: Publish bounded scanner summary
+      if: always()
+      shell: bash
+      run: '"$GITHUB_ACTION_PATH/scan.sh" summary'
+    - name: Retain complete scanner evidence
+      if: always()
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
+      with:
+        name: repository-security-reports
```
dceoy: **P2 — Make evidence artifact names rerun-safe**

This upload always runs before the aggregate enforcement step. A failed first attempt therefore creates `repository-security-reports`; using **Re-run failed jobs** then executes the same job in the same workflow run and attempts to create that immutable artifact name again. Because `overwrite` is not enabled, the upload can fail with a name conflict even when the transient scanner failure has cleared.

Prefer including `${{ github.run_attempt }}` in the artifact name so evidence from each attempt is retained independently (or explicitly use `overwrite: true`), and add a structural test for rerun-safe naming.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,560 @@
...
+  if ((scanner_status_value == 0 && render_status_value != 0)); then
+    scanner_status_value=${render_status_value}
+  fi
+  ensure_text_evidence actionlint "${scanner_status_value}"
+  printf '%s\n' "${scanner_status_value}" > "${results_dir}/actionlint.status"
+}
+
+run_shellcheck() {
+  local scanner_status_value render_status_value entry mode file first_line
+  local shell_shebang='^#![[:space:]]*(/usr/bin/env[[:space:]]+(-[[:alnum:]-]+[[:space:]]+)*)?(/[^[:space:]]*/)?(busybox[[:space:]]+)?(sh|ash|dash|bash|ksh93|ksh88|ksh|oksh|bats)([[:space:]]|$)'
```
dceoy: **P2 — Detect valid `env -S` shebangs with option operands**

The regex accepts repeated option tokens, but not options that require a following operand or environment assignments. Valid extensionless scripts such as `#!/usr/bin/env -S -u FOO bash` or `#!/usr/bin/env -S FOO=bar bash` do not match and are omitted from `script_files`, allowing them to bypass standalone ShellCheck while the gate reports success.

Parse the `env -S` argument string more robustly, or conservatively classify supported-shell shebangs, and add fixtures covering an option operand and an assignment.

dceoy (commented):
Two remaining issues affect rerun reliability and standalone ShellCheck coverage. The trust-boundary fixes and existing review feedback otherwise look resolved, and the latest CI run is green.

.github/actions/repository-security-scan/action.yml
dceoy: Fixed in 4f51685: evidence artifacts now include github.run_attempt, with a structural regression assertion.

.github/actions/repository-security-scan/scan.sh
dceoy: Fixed in 4f51685: the detector now conservatively scans supported shells after env arguments, with independent option-operand and assignment fixtures.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,560 @@
...
+  local file match
+  local directive_pattern='^[[:space:]]*#[[:space:]]*shellcheck[[:space:]]+[[:alpha:]-]+='
+  local directive_report="${results_dir}/${scanner_name}.shellcheck-directives"
+  local run_blocks="${results_dir}/${scanner_name}.run-blocks"
+  local matches="${results_dir}/${scanner_name}.shellcheck-matches"
+
+  : > "${directive_report}"
+  : > "${results_dir}/${scanner_name}.log"
+  for file in "$@"; do
+    if ! yq eval '.jobs[]?.steps[]? | select(has("run")) | .run' "${file}" \
```
dceoy, line 217: **[MEDIUM | high confidence] Composite action `run:` blocks bypass the trusted ShellCheck policy**

This extractor only inspects `.jobs[].steps[].run` in workflow files. `actionlint` does not lint composite action metadata, and the standalone ShellCheck gate only collects tracked script files, so shell code under `runs.steps[].run` in any local `action.yml`/`action.yaml` is never passed to ShellCheck. A target can therefore keep ShellCheck findings—or a `# shellcheck disable=...` directive—inside a composite action while this gate succeeds.

Collect composite-action run blocks as well and run them with the trusted ShellCheck configuration (including directive rejection), then add a fixture proving a finding in `runs.steps[].run` fails the gate.

dceoy (commented):
One actionable ShellCheck coverage gap remains. The latest CI run is green and all prior review threads are resolved, but composite-action inline shell is not currently covered by either actionlint or the standalone script collector.

.github/workflows/repository-security-scan.yml
```diff
@@ -0,0 +1,46 @@
...
+          path: _trusted
+          persist-credentials: false
+      - name: Check out untrusted target revision
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          # A merge queue must scan its synthetic head commit. Reusable calls
+          # from other events fall back to the exact triggering commit.
+          ref: ${{ github.event_name == 'merge_group' && github.event.merge_group.head_sha || github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
```
dceoy: **P1: Scan the pull-request merge result, not only the head commit**

For a `pull_request` run, `github.sha` is the synthetic merge commit, while `github.event.pull_request.head.sha` contains only the contributor branch. A stale branch can therefore pass this repository-wide gate even when the effective merge with the current base introduces a zizmor/actionlint, IaC, secret, or dependency finding. This is especially problematic when the check is used without a merge queue or an explicit up-to-date-branch requirement.

Use the pull-request merge SHA (`github.sha`) for the target checkout, while retaining `merge_group.head_sha` for merge queues. Add a fixture/structural assertion that the PR path scans the merge result.

.github/workflows/repository-security-scan.yml
```diff
@@ -0,0 +1,46 @@
...
+    steps:
+      - name: Check out trusted scanner implementation
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          repository: ${{ job.workflow_repository }}
+          # A pull request against this source repository must use its base
+          # revision as the scanner. Reusable and ruleset runs use the exact
+          # repository and SHA of the workflow that defines this job, not the
+          # caller workflow identity from the github context.
+          ref: ${{ job.workflow_repository == github.repository && github.event_name == 'pull_request' && github.event.pull_request.base.sha || job.workflow_sha }}
```
dceoy: **P1: Do not trust `job.workflow_sha` for a direct merge-group run**

Once #867 adds a direct `merge_group` trigger in this repository, this condition falls through to `job.workflow_sha`. For a direct workflow run, that workflow identity is resolved on the merge-group revision, so a queued PR can modify `.github/actions/repository-security-scan` or its trusted configuration and have the synthetic merge commit checked out as `_trusted`.

Handle same-repository `merge_group` events like pull requests and pin `_trusted` to `github.event.merge_group.base_sha`; reserve `job.workflow_sha` for reusable calls where the workflow reference itself is the reviewed pin. Add a boundary test before enabling the direct trigger.

.github/workflows/repository-security-scan.yml
```diff
@@ -0,0 +1,46 @@
...
+          path: _trusted
+          persist-credentials: false
+      - name: Check out untrusted target revision
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
```
dceoy, line 37: **P1: Reject or materialize Git LFS and submodule content before reporting success**

`actions/checkout` leaves both `lfs` and `submodules` disabled by default. Consequently an LFS-managed lockfile/secret is scanned only as a small pointer file, and a gitlink (`160000`) is left without its referenced tree; the preflight and collectors silently skip that content. A PR can therefore move scanner-relevant content behind LFS or a submodule and receive a green repository-wide gate.

Either materialize these objects under a tightly controlled policy, or more safely fail preflight on LFS pointers and gitlink index entries. Add fixtures for a tracked LFS pointer and a submodule entry so incomplete checkouts cannot pass.

dceoy (commented):
Three trust/coverage gaps remain. CI is green and the previously reported scanner-policy bypasses are resolved, but these cases can still let the required gate approve content other than the effective merged repository or omit tracked repository content entirely.

.github/workflows/repository-security-scan.yml
dceoy: Fixed in 537d82b: pull-request targets now use github.sha (the synthetic merge result), while merge-group targets retain head_sha, with a structural regression assertion.

.github/workflows/repository-security-scan.yml
dceoy: Fixed in 537d82b: preflight now rejects Git LFS pointer headers and 160000 submodule gitlinks before scanners run, with regression coverage for both cases.

.github/workflows/repository-security-scan.yml
dceoy: Fixed in 537d82b: direct same-repository merge-group runs now pin the trusted checkout to merge_group.base_sha, reserving job.workflow_sha for reusable calls.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,656 @@
...
+  local -n composite_output_ref=${output_name}
+  local action_file block_json block_index decode_status extracted_file
+  local blocks_jsonl="${results_dir}/shellcheck-composite-blocks.jsonl"
+  local extraction_dir="${results_dir}/shellcheck-composite"
+  local shell_pattern='(^|[[:space:]/])(sh|ash|dash|bash|ksh93|ksh88|ksh|oksh|bats)([[:space:]]|$)'
+
+  composite_output_ref=()
+  for action_file in "$@"; do
+    if ! yq eval --output-format=json --unwrapScalar=false \
+      "select(.runs.using == \"composite\") | .runs.steps[]? | select(has(\"run\")) | select(.shell | test(\"${shell_pattern}\")) | .run" \
```
dceoy, line 299: **P1 — Do not let expression-valued composite shells bypass ShellCheck**

This filter evaluates the literal metadata string, but composite-action `shell` is an expression-capable field (including the `inputs` context). A valid target action can therefore use:

```yaml
inputs:
  runner-shell:
    default: bash
runs:
  using: composite
  steps:
    - shell: ${{ inputs.runner-shell }}
      run: echo $unquoted
```

At runtime this executes with Bash, while the literal `${{ inputs.runner-shell }}` does not match `shell_pattern`. The run block is omitted from `composite_shell_files`, so both the ShellCheck finding and any suppression directive in it bypass this gate. Fail closed on dynamic shell expressions, or conservatively inspect every composite `run` block that is not provably a non-shell interpreter, and add an input-backed `bash` regression fixture.

.github/workflows/repository-security-scan.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Check out trusted scanner implementation
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          repository: ${{ job.workflow_repository }}
+          # A direct pull request or merge-group run against this source
+          # repository must use its base revision as the scanner. Reusable and
+          # ruleset runs use the exact repository and SHA of the workflow that
+          # defines this job, not the caller workflow identity from the github
+          # context.
+          ref: ${{ job.workflow_repository == github.repository && github.event_name == 'pull_request' && github.event.pull_request.base.sha || job.workflow_repository == github.repository && github.event_name == 'merge_group' && github.event.merge_group.base_sha || job.workflow_sha }}
```
dceoy: **P2 — Preserve the selected workflow SHA for same-repository calls**

Repository equality does not distinguish a direct workflow run from a reusable or ruleset workflow whose source repository also happens to be the target. In those latter cases `job.workflow_sha` is the reviewed commit defining this job, but this branch replaces it with the PR base SHA. A same-repository caller pinned to commit X can therefore execute scanner code and policy from a different base commit Y; the same drift applies when this repository is targeted by a ruleset sourced from itself.

Use workflow identity rather than repository equality to detect a genuinely direct run, and retain `job.workflow_sha` for reusable/ruleset executions. Add a structural test for a same-repository pinned call.

dceoy (commented):
Two remaining trust-boundary gaps can let target-controlled content bypass the intended pinned scanner policy. Please address the inline findings before merging.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -0,0 +1,681 @@
...
+  local -a workflow_files=()
+
+  preflight_succeeded actionlint || return 0
+  cd "${target_dir}" || return 0
+  collect_regular_files workflow_files workflows
+  if ((${#workflow_files[@]} == 0)); then
+    write_skip_evidence actionlint 'no tracked GitHub Actions workflow files were found.'
+    return 0
+  fi
+  if reject_workflow_shellcheck_directives actionlint "${workflow_files[@]}"; then
```
dceoy, line 391: **P1 — Reject expression-valued workflow shells before actionlint**

This precheck rejects ShellCheck directives, but it does not reject dynamic `shell` values in workflow steps or `defaults.run.shell`. A valid workflow can use, for example, `shell: ${{ matrix.shell }}` with `matrix.shell: [bash]`. At runtime the block executes as Bash, but actionlint v1.7.12 passes the literal expression to `getShellName`; its ShellCheck rule only accepts literal `bash`, `sh`, or strings beginning with those names, and otherwise returns without checking the script. A PR can therefore hide both ShellCheck findings and suppression directives in such a `run:` block while this gate succeeds.

Fail closed on expression-valued effective shells in workflows, as already done for composite actions, or independently extract and inspect those run blocks. Add a regression fixture with a matrix/input-backed Bash shell and a ShellCheck finding.

dceoy (commented):
One blocking ShellCheck bypass remains in workflow `run:` blocks. Expression-valued shells can execute as Bash at runtime while actionlint skips its ShellCheck integration, so the trusted gate can report success without checking that script.

dceoy (commented):
Reviewed current head `19d87e8`.

No additional blocking findings. The previously identified bootstrap, trusted-revision selection, incomplete-checkout, scanner-suppression, ShellCheck coverage, dynamic-shell, and artifact rerun issues are addressed, and the current review threads are resolved. The staged `workflow_call`-only rollout is internally consistent with deferring direct `pull_request`/`merge_group` activation to #867.

Current CI/CD run is green.

Issue #867, opened 2026-07-31
Activate direct triggers and support immutable target scans


dceoy:
## Current status

- [x] #866 merged on 2026-08-02 as `1413c89221edd788de6dcbcb30f7e6973615048f`.
- [x] `main` now contains the trusted scanner implementation at `.github/actions/repository-security-scan`.
- [ ] Direct `pull_request` and `merge_group` triggers remain disabled in `.github/workflows/repository-security-scan.yml`.
- [ ] The reusable workflow can scan only the caller's triggering repository/revision; it cannot yet execute a periodic scan of an explicitly supplied immutable repository commit for `dceoy/segh`.

## Background

#866 added the reusable, read-only `Repository security` workflow, its composite action, scanner script, centrally owned configuration, tests, and Aqua pins. The bootstrap revision intentionally exposed only `workflow_call`: a direct pull-request or merge-group run would have selected its base revision as the trusted scanner, but no prior base revision contained the scanner action.

That bootstrap constraint is now resolved because the implementation is reachable from protected `main`.

`dceoy/segh` also reuses the same scanner implementation for periodic organization-wide default-branch scans. It currently duplicates repository checkout, token scoping, scanner invocation, evidence validation, result classification, and artifact publication in a large local workflow. Extend the reusable workflow with a narrowly defined immutable-target mode so those generic scanner responsibilities can move to `gha-for-devops` without allowing the target repository to control the scanner or policy.

## Scope 1: activate direct required-workflow triggers

1. Add `pull_request:` and `merge_group:` to `on:` in `.github/workflows/repository-security-scan.yml` while retaining `workflow_call:`.
2. Keep direct pull-request runs pinned to `github.event.pull_request.base.sha` for the trusted scanner and scan the pull request's synthetic merge commit.
3. Keep direct merge-group runs pinned to `github.event.merge_group.base_sha` for the trusted scanner and scan `github.event.merge_group.head_sha`.
4. Preserve the stable aggregate check name `Repository security / scan`.
5. Update `.github/scripts/tests/repository-security-scan.bats` to pin the direct triggers and trusted/target revision selection.
6. Update `README.md.tmpl` and `README.md` to remove the bootstrap caveat and document direct organization-ruleset use.

## Scope 2: add explicit immutable-target mode for periodic scans

Extend `workflow_call` with the smallest interface needed to scan a repository selected by a trusted organization controller such as `dceoy/segh`.

Suggested inputs:

[...]

PR #868, opened 2026-08-03
Support immutable target scans for repository security workflow

dceoy:
## Summary

Progresses #867 — this PR lands Scope 2 (explicit immutable-target mode). Scope 1 (direct `pull_request`/`merge_group` trigger activation) remains open; see below.

**Scope 1 — activate direct required-workflow triggers (deferred, not landed in this PR)**
- Direct `pull_request`/`merge_group` triggers were added and live-validated via this PR's own `Repository security / scan` check, which surfaced a pre-existing, unrelated blocker: target-owned ShellCheck `shell=`/`source=`/`external-sources=` directives (load-bearing for `scripts/qa.sh`) fail the base-pinned trusted scanner's directive policy.
- `on:` has been reverted to `workflow_call:` only (commit `6b1125a`) until that directive-policy question is resolved on `main`; `README.md.tmpl`/`README.md` still document the bootstrap caveat accordingly.
- Activating direct triggers is tracked as follow-up work under #867.

**Scope 2 — explicit immutable-target mode for periodic scans**
- `workflow_call` gains `target-repository`, `target-ref`, `repository-id`, `default-branch`, `evidence-artifact-name` inputs and optional `TARGET_APP_ID`/`TARGET_APP_PRIVATE_KEY` secrets.
- A new `.github/actions/repository-security-scan/resolve-target.sh` validates/resolves the scan target (rejects a target repo without a full 40-character lowercase commit SHA, or a `target-ref` without `target-repository`) and is the single source for the checkout `repository`/`ref` used by both the target checkout step and the composite action call.
- The trusted scanner revision is derived only from `job.workflow_repository`/`job.workflow_sha`, exactly as before — it can never be influenced by `target-repository`.
- A short-lived, repository-scoped GitHub App token (`contents: read`, `metadata: read` only) is minted via `actions/create-github-app-token` only when `target-repository` is set; direct triggers and same-repo `workflow_call` are unaffected and keep using `github.token`.
- The target checkout explicitly disables `persist-credentials`, `lfs`, and `submodules`.
- The composite action now emits a compact `status.json` (`pass`/`findings`/`incomplete`/`error`) bound to `repository-id`/`repository`/`default-branch`/`commit-sha`, only when `repository-id` or `default-branch` is supplied, for controller-side aggregation across matrix invocations. It also accepts `evidence-artifact-name` so matrix callers can publish uniquely named artifacts.
- Added a `concurrency:` group (keyed by PR number / merge-group head ref / `target-repository` / run ID) — this addresses a zizmor pedantic-persona finding that only appears now that the workflow has its own triggers; `target-repository` in the key keeps concurrent matrix invocations from a controller from colliding.

## Verification limits (please read before merging)

- Scope 2's own token-minting, cross-repo checkout, and matrix-aggregation behavior **cannot be live-verified from this environment** — there's no GitHub App configured, no private test repository, and no access to `dceoy/segh`. Everything there is validated statically (bats, actionlint, zizmor) and by code review, not by an end-to-end run.
- The self-scan this PR triggers exercises the *content* of the changed files (via the unchanged trusted scanner from `main`), not the *new* `status` step, `evidence-artifact-name`, or classification logic at runtime — those are covered by new bats tests only (`.github/scripts/tests/repository-security-scan.bats`, 15 new cases).
- Downstream rollout steps from the issue (pinning the new SHA in the org ruleset, verifying on `dceoy/segh` and other repos, migrating `segh`'s periodic worker, closing #865) are out of scope for this PR and listed as follow-ups in #867.

## Test plan

- [x] `bats .github/scripts/tests/repository-security-scan.bats` (44 tests; 5 pre-existing failures are macOS `sed -i` incompatibilities unrelated to this change and confirmed present on `main` too — they pass on Linux CI)
- [x] `shellcheck --severity=style --rcfile .github/security/repository-security/shellcheckrc` on `scan.sh` and the new `resolve-target.sh`
- [x] `actionlint --config-file .github/security/repository-security/actionlint.yaml --shellcheck shellcheck` on the workflow
- [x] `zizmor --offline --no-config --no-ignores --strict-collection --persona regular --min-severity medium --min-confidence high` — no findings (also clean under `--persona pedantic`)
- [x] `scripts/qa.sh` (repo-wide shfmt/shellcheck/actionlint/yamllint/zizmor/checkov) — no findings, no unexpected file changes
- [x] `.github/scripts/update-readme.sh` — regenerated `README.md` reviewed
- [ ] Live verification of the explicit immutable-target path (GitHub App token scoping, private-repo checkout, matrix artifact aggregation) — needs App credentials and a test repository outside this environment

### DIFF
--- a/.github/actions/repository-security-scan/action.yml
+++ b/.github/actions/repository-security-scan/action.yml
@@ -1,6 +1,29 @@
 ---
 name: Trusted repository security scanner pipeline
 description: Run centrally configured security gates against an untrusted repository checkout
+inputs:
+  repository-id:
+    description: Stable repository identity copied into periodic-scan status evidence
+    required: false
+    default: ''
+  default-branch:
+    description: Default branch identity copied into periodic-scan status evidence
+    required: false
+    default: ''
+  repository:
+    description: Repository (owner/name) actually scanned, copied into status evidence
+    required: false
+    default: ''
+  commit-sha:
+    description: Commit SHA actually scanned, copied into status evidence
+    required: false
+    default: ''
+  evidence-artifact-name:
+    description: >-
+      Unique evidence artifact name for matrix callers. Empty uses a
+      run-attempt-scoped name.
+    required: false
+    default: ''
 runs:
   using: composite
   steps:
@@ -41,6 +64,15 @@ runs:
       if: always()
       shell: bash
       run: '"$GITHUB_ACTION_PATH/scan.sh" trivy-secret'
+    - name: Record periodic-scan status evidence
+      if: always()
+      shell: bash
+      env:
+        SECURITY_SCAN_REPOSITORY_ID: ${{ inputs.repository-id }}
+        SECURITY_SCAN_DEFAULT_BRANCH: ${{ inputs.default-branch }}
+        SECURITY_SCAN_REPOSITORY: ${{ inputs.repository }}
+        SECURITY_SCAN_COMMIT_SHA: ${{ inputs.commit-sha }}
+      run: '"$GITHUB_ACTION_PATH/scan.sh" status'
     - name: Publish bounded scanner summary
       if: always()
       shell: bash
@@ -49,7 +81,7 @@ runs:
       if: always()
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
       with:
-        name: repository-security-reports-${{ github.run_attempt }}
+        name: ${{ inputs.evidence-artifact-name != '' && inputs.evidence-artifact-name || format('repository-security-reports-{0}', github.run_attempt) }}
         path: security-results
         if-no-files-found: error
         retention-days: 14
--- a/.github/actions/repository-security-scan/action.yml
+++ b/.github/actions/repository-security-scan/action.yml
@@ -24,6 +24,13 @@ inputs:
       run-attempt-scoped name.
     required: false
     default: ''
+  setup-failed:
+    description: >-
+      Set by the caller when target resolution, token minting, or checkout
+      failed before this composite ran, so preflight records explicit error
+      evidence instead of inspecting an absent or incomplete checkout.
+    required: false
+    default: 'false'
 runs:
   using: composite
   steps:
@@ -39,6 +46,8 @@ runs:
     - name: Validate the target checkout
       if: always()
       shell: bash
+      env:
+        SECURITY_SCAN_SETUP_FAILED: ${{ inputs.setup-failed }}
       run: '"$GITHUB_ACTION_PATH/scan.sh" preflight'
     - name: Run zizmor gate
       if: always()
--- a/.github/actions/repository-security-scan/action.yml
+++ b/.github/actions/repository-security-scan/action.yml
@@ -38,8 +38,9 @@ runs:
       # Must run even when the calling job's earlier setup steps (target
       # resolution, token minting, checkout) failed: without an explicit
       # always-run condition, this step's implicit success() is skipped in
-      # that case, and the later always-run status step's yq dependency is
-      # then missing, losing the promised failure-path status evidence.
+      # that case, preventing the scanner pipeline from using its trusted
+      # toolchain when preflight can proceed.
+      id: install-scanners
       if: (! cancelled())
       uses: aquaproj/aqua-installer@96a9bc20066c5bf5e275b41019cfc165b25f4e2e  # v4.0.5
       with:
@@ -50,10 +51,15 @@ runs:
       shell: bash
       run: '"$GITHUB_ACTION_PATH/scan.sh" prepare'
     - name: Validate the target checkout
+      # Folds this step's own installer failure into the same setup-failed
+      # signal as an upstream checkout failure: the scanner toolchain
+      # (including yq) is unavailable either way, so preflight must fail
+      # closed with explicit error evidence instead of letting each gate
+      # hit missing binaries on its own.
       if: always()
       shell: bash
       env:
-        SECURITY_SCAN_SETUP_FAILED: ${{ inputs.setup-failed }}
+        SECURITY_SCAN_SETUP_FAILED: ${{ inputs.setup-failed == 'true' || steps.install-scanners.outcome == 'failure' }}
       run: '"$GITHUB_ACTION_PATH/scan.sh" preflight'
     - name: Run zizmor gate
       if: always()
--- a/.github/actions/repository-security-scan/action.yml
+++ b/.github/actions/repository-security-scan/action.yml
@@ -35,6 +35,12 @@ runs:
   using: composite
   steps:
     - name: Install checksum-verified scanners
+      # Must run even when the calling job's earlier setup steps (target
+      # resolution, token minting, checkout) failed: without an explicit
+      # always-run condition, this step's implicit success() is skipped in
+      # that case, and the later always-run status step's yq dependency is
+      # then missing, losing the promised failure-path status evidence.
+      if: (! cancelled())
       uses: aquaproj/aqua-installer@96a9bc20066c5bf5e275b41019cfc165b25f4e2e  # v4.0.5
       with:
         aqua_version: v2.62.1
--- a/.github/actions/repository-security-scan/resolve-target.sh
+++ b/.github/actions/repository-security-scan/resolve-target.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -u -o pipefail
+
+main() {
+  local target_repository="${TARGET_REPOSITORY:-}"
+  local target_ref="${TARGET_REF:-}"
+
+  if [[ -n "${target_repository}" ]]; then
+    if [[ ! "${target_repository}" =~ ^[^/[:space:]]+/[^/[:space:]]+$ ]]; then
+      printf '::error::target-repository must be an owner/name pair.\n' >&2
+      return 1
+    fi
+    if [[ ! "${target_ref}" =~ ^[0-9a-f]{40}$ ]]; then
+      printf '::error::target-ref must be a full lowercase 40-character commit SHA when target-repository is set.\n' >&2
+      return 1
+    fi
+    {
+      printf 'owner=%s\n' "${target_repository%%/*}"
+      printf 'name=%s\n' "${target_repository#*/}"
+      printf 'repository=%s\n' "${target_repository}"
+      printf 'ref=%s\n' "${target_ref}"
+    } >> "${GITHUB_OUTPUT}"
+    return 0
+  fi
+
+  if [[ -n "${target_ref}" ]]; then
+    printf '::error::target-ref requires target-repository to be set.\n' >&2
+    return 1
+  fi
+
+  {
+    printf 'owner=\n'
+    printf 'name=\n'
+    printf 'repository=%s\n' "${EVENT_REPOSITORY:-}"
+    if [[ "${EVENT_NAME:-}" == merge_group ]]; then
+      printf 'ref=%s\n' "${MERGE_GROUP_HEAD_SHA:-}"
+    else
+      printf 'ref=%s\n' "${EVENT_SHA:-}"
+    fi
+  } >> "${GITHUB_OUTPUT}"
+}
+
+main
--- a/.github/actions/repository-security-scan/resolve-target.sh
+++ b/.github/actions/repository-security-scan/resolve-target.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -u -o pipefail
+set -euo pipefail
 
 main() {
   local target_repository="${TARGET_REPOSITORY:-}"
--- a/.github/actions/repository-security-scan/scan.sh
+++ b/.github/actions/repository-security-scan/scan.sh
@@ -422,7 +422,7 @@ run_actionlint() {
       > "${results_dir}/actionlint.json" 2>> "${results_dir}/actionlint.log"; then
       printf '::error::actionlint produced JSON Lines output that could not be converted to a JSON array; see actionlint.log.\n'
       printf '[]\n' > "${results_dir}/actionlint.json"
-      scanner_status_value=1
+      scanner_status_value=2
     fi
   else
     printf '[]\n' > "${results_dir}/actionlint.json"
--- a/.github/actions/repository-security-scan/scan.sh
+++ b/.github/actions/repository-security-scan/scan.sh
@@ -383,8 +383,13 @@ run_zizmor() {
     > "${results_dir}/zizmor.txt" \
     2>> "${results_dir}/zizmor.log"
   render_status_value=$?
-  if ((scanner_status_value == 0 && render_status_value != 0)); then
-    scanner_status_value=${render_status_value}
+  if ((render_status_value != 0)); then
+    if ((render_status_value == 1)); then
+      render_status_value=2
+    fi
+    if ((render_status_value > scanner_status_value)); then
+      scanner_status_value=${render_status_value}
+    fi
   fi
   ensure_text_evidence zizmor "${scanner_status_value}"
   printf '%s\n' "${scanner_status_value}" > "${results_dir}/zizmor.status"
@@ -438,7 +443,7 @@ run_actionlint() {
     > "${results_dir}/actionlint.txt" \
     2>> "${results_dir}/actionlint.log"
   render_status_value=$?
-  if ((scanner_status_value == 0 && render_status_value != 0)); then
+  if ((render_status_value > scanner_status_value)); then
     scanner_status_value=${render_status_value}
   fi
   ensure_text_evidence actionlint "${scanner_status_value}"
@@ -543,7 +548,7 @@ run_shellcheck() {
     2>> "${results_dir}/shellcheck.log"
   render_status_value=$?
   rm -rf -- "${composite_shell_dir}"
-  if ((scanner_status_value == 0 && render_status_value != 0)); then
+  if ((render_status_value > scanner_status_value)); then
     scanner_status_value=${render_status_value}
   fi
   ensure_text_evidence shellcheck "${scanner_status_value}"
@@ -631,8 +636,13 @@ run_trivy() {
     "${results_dir}/${scanner_name}.json" \
     2>> "${results_dir}/${scanner_name}.log"
   convert_status_value=$?
-  if ((scanner_status_value == 0 && convert_status_value != 0)); then
-    scanner_status_value=${convert_status_value}
+  if ((convert_status_value != 0)); then
+    if ((convert_status_value == 1)); then
+      convert_status_value=2
+    fi
+    if ((convert_status_value > scanner_status_value)); then
+      scanner_status_value=${convert_status_value}
+    fi
   fi
   ensure_text_evidence "${scanner_name}" "${scanner_status_value}"
   printf '%s\n' "${scanner_status_value}" > "${results_dir}/${scanner_name}.status"
--- a/.github/actions/repository-security-scan/scan.sh
+++ b/.github/actions/repository-security-scan/scan.sh
@@ -626,7 +626,7 @@ run_trivy() {
 
 classify_result() {
   local preflight_status_value='' scanner_name scanner_status_value
-  local has_incomplete=0 has_findings=0
+  local has_incomplete=0 has_findings=0 has_error=0
 
   if [[ ! -f "${results_dir}/preflight.status" ]]; then
     printf 'error'
@@ -671,12 +671,20 @@ classify_result() {
     if yq eval --exit-status --input-format=json '.result == "not-run"' "${scanner_json_file}" \
       > /dev/null 2>&1; then
       has_incomplete=1
+    elif ((scanner_status_value > 1)); then
+      # A status of 1 is every scanner's own convention for "policy findings
+      # at the enforced threshold". Anything higher (and not already claimed
+      # as not-run above) means the tool itself failed to complete its scan
+      # (e.g. a usage error or crash), which is not a policy finding.
+      has_error=1
     else
       has_findings=1
     fi
   done
 
-  if ((has_incomplete == 1)); then
+  if ((has_error == 1)); then
+    printf 'error'
+  elif ((has_incomplete == 1)); then
     printf 'incomplete'
   elif ((has_findings == 1)); then
     printf 'findings'
--- a/.github/actions/repository-security-scan/scan.sh
+++ b/.github/actions/repository-security-scan/scan.sh
@@ -634,13 +634,15 @@ classify_result() {
   fi
   preflight_status_value="$(< "${results_dir}/preflight.status")"
   if [[ "${preflight_status_value}" != '0' ]]; then
-    if [[ ! -f "${results_dir}/tracked-files.index" ]]; then
-      printf 'error'
-    else
-      # A non-zero preflight status with a tracked-files index means the
-      # target itself carried a rejected LFS pointer or submodule gitlink;
-      # that is a detected policy violation, not an execution failure.
+    if [[ -f "${results_dir}/rejected-lfs-pointers.txt" ]] \
+      || [[ -f "${results_dir}/rejected-submodules.txt" ]]; then
+      # These files are only written once preflight has fully enumerated the
+      # checkout, so their presence means the non-zero status came from a
+      # rejected LFS pointer or submodule gitlink, not a failed inspection
+      # (e.g. `git ls-files` itself erroring before enumeration completed).
       printf 'findings'
+    else
+      printf 'error'
     fi
     return 0
   fi
--- a/.github/actions/repository-security-scan/scan.sh
+++ b/.github/actions/repository-security-scan/scan.sh
@@ -674,6 +674,11 @@ classify_result() {
       has_incomplete=1
       continue
     fi
+    if [[ ! -s "${results_dir}/${scanner_name}.txt" ]] \
+      || [[ ! -f "${results_dir}/${scanner_name}.log" ]]; then
+      has_incomplete=1
+      continue
+    fi
     if ((scanner_status_value == 0)); then
       continue
     fi
@@ -702,6 +707,17 @@ classify_result() {
   fi
 }
 
+json_escape() {
+  local value="$1"
+
+  value="${value//\\/\\\\}"
+  value="${value//\"/\\\"}"
+  value="${value//$'\n'/\\n}"
+  value="${value//$'\r'/\\r}"
+  value="${value//$'\t'/\\t}"
+  printf '%s' "${value}"
+}
+
 record_status() {
   local repository_id="${SECURITY_SCAN_REPOSITORY_ID:-}"
   local default_branch="${SECURITY_SCAN_DEFAULT_BRANCH:-}"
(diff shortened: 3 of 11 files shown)


.github/workflows/repository-security-scan.yml
```diff
@@ -33,14 +76,51 @@ jobs:
...
+        id: target
+        shell: bash
+        env:
+          TARGET_REPOSITORY: ${{ inputs.target-repository }}
+          TARGET_REF: ${{ inputs.target-ref }}
+          EVENT_NAME: ${{ github.event_name }}
+          EVENT_REPOSITORY: ${{ github.repository }}
+          EVENT_SHA: ${{ github.sha }}
+          MERGE_GROUP_HEAD_SHA: ${{ github.event.merge_group.head_sha }}
+        run: '"${GITHUB_WORKSPACE}/_trusted/.github/actions/repository-security-scan/resolve-target.sh"'
```
dceoy, line 105: **P0 — Direct triggers cannot execute this newly added trusted helper**

For a direct `pull_request` run, the preceding checkout intentionally pins `_trusted` to `github.event.pull_request.base.sha`. That base is `1413c892...`, where `resolve-target.sh` does not exist, so the live run fails here with exit code 127 (`No such file or directory`) and never scans the target. The same bootstrap constraint removed from the header still applies to this new helper.

Keep the direct-trigger path limited to files already present on the base revision—for example, skip this helper when `target-repository` is empty and use event expressions for the normal repository/ref path—or land the helper separately before enabling direct triggers.

dceoy (commented):
Blocking finding: the direct-trigger path is currently broken. The `Repository security` run fails before checkout/scanning because the trusted base revision does not contain the newly added resolver. This defeats Scope 1 and should be fixed before merge. The two existing unresolved findings (`set -euo pipefail` and preflight error classification) should also be addressed.

.github/actions/repository-security-scan/resolve-target.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -u -o pipefail
```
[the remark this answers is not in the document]
dceoy: Fixed — `resolve-target.sh` now uses `set -euo pipefail`, verified by the existing bats coverage.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -624,6 +624,86 @@ run_trivy() {
...
+    printf 'error'
+    return 0
+  fi
+  preflight_status_value="$(< "${results_dir}/preflight.status")"
+  if [[ "${preflight_status_value}" != '0' ]]; then
+    if [[ ! -f "${results_dir}/tracked-files.index" ]]; then
+      printf 'error'
+    else
+      # A non-zero preflight status with a tracked-files index means the
+      # target itself carried a rejected LFS pointer or submodule gitlink;
```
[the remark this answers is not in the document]
dceoy: Fixed — `classify_result` now keys off `rejected-lfs-pointers.txt`/`rejected-submodules.txt` (only written once preflight fully enumerates the checkout) instead of `tracked-files.index`'s mere existence, so a failed `git ls-files` now reports `error` instead of `findings`. Added a bats regression test for this exact scenario and confirmed it fails against the old code.

.github/workflows/repository-security-scan.yml
dceoy: Fixed the exit-127 by skipping this helper (falling back to plain event expressions) when `target-repository` is empty — see the PR-level reply for a new, unreviewed blocker this uncovered and a question on how you'd like to proceed.

dceoy:
The exit-127 from `resolve-target.sh` missing on the base checkout is fixed by skipping that step (falling back to plain event expressions) when `target-repository` is empty. That gets the direct-trigger path past step 1, but the real `scan` run then hits a different, previously-unreviewed blocker: the target checkout is this repo's own tree, and `.github/scripts/tests/fixtures/repository-security-scan/lfs-pointer/large-lockfile.fixture` is itself a committed file with real LFS-pointer-pattern content, so preflight rejects it and every scanner gate reports not-run. None of the scanners or preflight checks currently exclude `.github/scripts/tests/fixtures/`, and since this PR's own `scan` job runs the composite action pinned to `_trusted`@base, no fix to `scan.sh` can be validated by this PR's own check before merge anyway.

Given that, your second option — landing the helper (and now also a fixtures fix) separately before enabling direct triggers — looks safer than patching around it here. Want me to (a) defuse the fixture's LFS-pointer content the same way `prepare_fixture` already defuses fixture shebangs (`#!fixture ` → `#!`), (b) add explicit skip-paths for the fixtures directory across scanners/preflight, or (c) split this PR so direct triggers land only once main already carries all of this?

dceoy (commented):
Reviewed current head `2a0817fbc70c`.

Blocking findings remain:

1. **P0 — The newly activated direct check is still unusable.** The current `Repository security / scan` run (`30820844821`) fails because the repository itself tracks `.github/scripts/tests/fixtures/repository-security-scan/lfs-pointer/large-lockfile.fixture` with literal Git LFS pointer content. The base-pinned scanner correctly rejects it, so this PR does not satisfy the live direct-trigger validation and the workflow would continue failing on this repository after merge. Prefer defusing the stored fixture and reconstructing the real pointer in `prepare_fixture`; do not exclude the fixtures directory from security scanning.

2. **P1 — Scanner execution errors can be recorded as policy findings.** `classify_result` treats every nonzero numeric status with valid JSON as `findings` unless the JSON has `.result == "not-run"`. For example, if `actionlint` exits `2` without diagnostics, `run_actionlint` writes `[]` plus status `2`, which is then classified as `findings` rather than `error`. The same issue applies to other scanners and to post-enumeration preflight errors such as failure to remove a symlink. Add an explicit error path (normally status `> 1`, with any tool-specific handling) and regression tests for scanner runtime failure.

3. **P2 — The target-ref-only rejection is unreachable in the workflow.** `resolve-target.sh` tests that `target-ref` without `target-repository` is rejected, but the step runs only when `inputs.target-repository != ''`; therefore a caller supplying only `target-ref` silently scans `github.sha`. Either invoke validation when either target input is present, or remove the dead rejection/test and document the intentional ignore consistently.

CI/CD passes, but the required security workflow is red. I do not recommend merging until P0 and P1 are resolved.

.github/workflows/repository-security-scan.yml
dceoy: Exit-127 is fixed on head 2a0817f, but the live check still can't go green from this PR alone — see the PR-level comment for what's still blocking.

dceoy:
Pushed 7ea2de3 against the 14:21 review. Status:

- **P1 (`classify_result` misclassifying scanner errors as findings)** — fixed: a status `> 1` without `not-run` evidence now reports `error`, not `findings`. Added a bats regression test (`actionlint-runtime-error`).
- **P2 (`target-ref`-only silently ignored)** — fixed: `Resolve immutable scan target` now also runs when `target-ref` is set alone, so `resolve-target.sh` rejects that combination instead of silently falling back to the event ref. Updated the input description to match.
- **P0 (fixture LFS pointer)** — fixed the reported layer, and while verifying with a *real* (non-stub) run of `scan.sh` against this repo's own tree, found a second undefused fixture (`shellcheck-directive/scripts/tool.sh.fixture` had a real, non-`#!fixture` shebang) that was also leaking a real ShellCheck-disable directive into the self-scan. Fixed the same way.

With both fixture layers fixed, the run on this head (job [30823943873](https://github.com/dceoy/gha-for-devops/actions/runs/30823943873/job/91720609564)) gets past preflight (`0`) and zizmor/checkov/trivy-vulnerability/trivy-secret all pass (`0`). **actionlint and shellcheck still fail** — not on any fixture, but on 11 pre-existing, unrelated ShellCheck directives across 9 files this PR doesn't touch:

- `.github/workflows/github-actions-lint-and-scan.yml:2`, `joern-scan.yml:6,19`, `json-schema-validation.yml:19`, `shell-project-ci.yml:55`
- `.agents/skills/local-qa/scripts/qa.sh:9`, `supply-chain-cooldown.sh:1`, `.github/scripts/tests/go-package-test.bats:44,207`, `pnpm-version-inference.bats:33`, `shell-project-ci.bats:49`

These directives are load-bearing for this repo's own `scripts/qa.sh` (e.g. removing `# shellcheck shell=bash` from `supply-chain-cooldown.sh` turns up a real SC2148, since that file is sourced rather than executed). `reject_workflow_shellcheck_directives`/`reject_script_shellcheck_directives` are unchanged from `main` — this isn't a regression from this PR, it's the first time anything has ever run these gates against this repo's real tree, because the LFS-pointer block was masking them completely.

So the direct-trigger self-scan still can't go green from this PR alone. Two ways I can see to unblock it, both your call:
1. Land the fixture fixes + P1 + P2 now, and defer trigger activation (your original option (c)) until the directive conflict is resolved separately.
2. Narrow `reject_*_shellcheck_directives`'s pattern to suppression directives (`disable=`) only, leaving semantic ones (`shell=`, `source=`, `external-sources=`) alone — happy to make that change here if you'd rather.

One more limitation worth flagging: `trivy`'s `--exit-code 1` means "findings" and "trivy itself failed" are the same exit code, so a hard trivy runtime failure would land on `incomplete` (via the existing empty/invalid-JSON check) rather than the new `error` path — not silently mis-tagged as `findings`, but not distinguishable as `error` either.

To be clear about what the green parts of this run prove: they exercise target *content* through the base-pinned trusted scanner (`_trusted`@main), not this PR's own `classify_result` change or the widened workflow `if` — those only take effect once this PR's `scan.sh`/workflow become the trusted revision itself.

dceoy (commented):
Reviewed current head `7ea2de3027c`.

Blocking findings remain:

1. **P1 — Target setup failures never produce the promised status/evidence.** `Resolve immutable scan target`, App-token minting, and target checkout all run before the composite scanner step. If the ref is malformed, App credentials are missing/invalid, the repository is inaccessible, or checkout fails, GitHub skips `Run trusted repository security pipeline`; therefore `status.json` is not created and no evidence artifact is uploaded. This contradicts #867's requirement to classify malformed/inaccessible targets and retain evidence on failure. Add an always-run finalization path (or invoke the composite with `always()` and pass explicit setup-failure state) so these cases emit `error` evidence.

2. **P1 — A symlink-removal runtime failure is still classified as a policy finding.** `run_preflight` always writes `rejected-lfs-pointers.txt` and `rejected-submodules.txt` after successful enumeration, including the “No tracked …” case. `classify_result` checks only whether either file exists. Consequently, if removing a tracked symlink fails, `preflight.status` becomes `1`, both unrelated files exist, and the result is reported as `findings` instead of `error`. Record explicit preflight finding/error evidence or inspect actual rejection counts/content, and add a regression test for failed symlink removal.

3. **P2 — The newly introduced App credential interface uses a deprecated action input.** The pinned `actions/create-github-app-token` revision marks `app-id` as deprecated in favor of `client-id`. Prefer `TARGET_APP_CLIENT_ID` / `client-id` rather than adding a new public interface around the deprecated field.

Verification note: `CI/CD` passes, but the current `Repository security` run (`30823943873`) is red because the base-pinned scanner rejects existing target-owned ShellCheck directives. The direct trigger is executing, but this PR currently validates only the findings path, not the representative clean-pass path required by #867.

I do not recommend merging until the two P1 classification/evidence gaps are fixed.

.github/workflows/repository-security-scan.yml
dceoy: Exit-127 was fixed on 2a0817f. The follow-on blocker (target-owned ShellCheck `shell=`/`source=`/`external-sources=` directives failing this repo's own self-scan) is resolved by deferring trigger activation in 6b1125a — `on:` is back to `workflow_call` only, with every target-mode input/secret/fix from this PR kept intact.

dceoy:
Pushed 377ecea against the 14:53 review, then 6b1125a to defer trigger activation:

- **P1 (target setup failures skip the composite, losing evidence)** — fixed: `Run trusted repository security pipeline` now runs `if: (! cancelled())` (not `always()`, since `pull_request` runs cancel a superseded in-progress scan) and threads a `setup-failed` flag into `scan.sh preflight`, so a failed resolve/token-mint/checkout now records explicit `error` evidence instead of the job going red with nothing to classify or upload.
- **P1 (symlink-removal failure misclassified as `findings`)** — fixed: `run_preflight` now writes an explicit `preflight.result` (`findings`/`error`) instead of `classify_result` inferring it from whether `rejected-lfs-pointers.txt`/`rejected-submodules.txt` exist, which was true unconditionally once enumeration completed, regardless of cause.
- **P2 (deprecated `app-id`)** — fixed: renamed the still-unreleased `TARGET_APP_ID` secret to `TARGET_APP_CLIENT_ID` and switched to `client-id`.

Since the live check still can't go green from this PR alone — target-owned `shell=`/`source=`/`external-sources=` directives (not a suppression mechanism) are pre-existing and load-bearing for `scripts/qa.sh` — I've gone with option 1: reverted `on:` to `workflow_call` only, keeping every explicit immutable-target input/secret and this round's fixes in place. Direct-trigger activation (this PR's original title scope) is now a follow-up once the directive-policy question is resolved on main.

Bats (50/50 in this suite, 124/124 overall), actionlint, and `scripts/qa.sh` all pass locally.

.github/actions/repository-security-scan/action.yml
```diff
@@ -16,6 +46,8 @@ runs:
     - name: Validate the target checkout
       if: always()
       shell: bash
+      env:
+        SECURITY_SCAN_SETUP_FAILED: ${{ inputs.setup-failed }}
```
dceoy: **P1 — Ensure the failure finalizer installs its dependencies**

The workflow now enters this composite with `if: (! cancelled())` after resolve/token/checkout failures, but the Aqua installer remains guarded by the implicit `success()` condition. Once an earlier job step has failed, that first composite step can be skipped. The later always-run `status` step calls `yq`; without the install, `status.json` generation fails and the malformed/inaccessible-target path again loses the promised status evidence. Add `if: (! cancelled())` to the installer (or make the failure finalizer independent of Aqua/yq).

dceoy (commented):
Reviewed current head `6b1125a78307`.

A blocking failure-path finding is attached inline. CI/CD is green, but the new immutable-target path is not yet reliable for malformed/inaccessible-target cases.

The PR metadata also remains stale: the title/body still claim direct-trigger activation and `Closes #867`, while this head deliberately restores `workflow_call` only. Since #867 explicitly requires live `pull_request`/`merge_group` triggers, update the title/body and remove `Closes #867` (or split the remaining direct-trigger work into a separately linked issue) before merge.

.github/workflows/repository-security-scan.yml
```diff
@@ -33,14 +81,72 @@ jobs:
...
+        # nothing to classify or upload. Unlike always(), this must not run
+        # on cancellation: pull_request runs cancel a superseded in-progress
+        # scan, and always() would start the full scanner pipeline against
+        # its 45-minute timeout instead of aborting.
+        if: (! cancelled())
         uses: ./_trusted/.github/actions/repository-security-scan
+        with:
+          repository-id: ${{ inputs.repository-id }}
+          default-branch: ${{ inputs.default-branch }}
+          repository: ${{ inputs.target-repository != '' && steps.target.outputs.repository || github.repository }}
```
dceoy: **P1 — Preserve the requested target identity when resolution fails**

For the exact failures this finalizer handles (invalid `target-repository` or malformed `target-ref`), `steps.target.outputs.repository/ref` are unset. These `&& ... || ...` expressions then fall back to `github.repository` and `github.sha`, so the resulting `error` status is bound to the caller rather than the requested target. Controller aggregation can therefore attribute a failed target scan to the wrong repository/commit. Use the raw `inputs.target-repository`/`inputs.target-ref` for evidence identity when target mode was requested; keep resolver outputs only for checkout/token operations.

dceoy (commented):
Additional blocking identity issue on the same head.

.github/actions/repository-security-scan/action.yml
dceoy: Fixed in 5b71886 — the installer step now has `if: (! cancelled())`.

.github/workflows/repository-security-scan.yml
dceoy: Fixed in 5b71886 — evidence repository/commit-sha now use inputs.target-repository/inputs.target-ref directly, with target-ref last in the expression so an empty ref surfaces as empty evidence rather than falling back to github.sha.

dceoy:
Title/body updated to drop "Closes #867" and reflect that Scope 1 (direct trigger activation) is deferred to a follow-up under #867, per the 15:23 review.

.github/actions/repository-security-scan/scan.sh
```diff
@@ -624,6 +638,91 @@ run_trivy() {
...
+    if [[ ! -f "${scanner_status_file}" ]]; then
+      has_incomplete=1
+      continue
+    fi
+    scanner_status_value="$(< "${scanner_status_file}")"
+    if [[ ! "${scanner_status_value}" =~ ^[0-9]+$ ]]; then
+      has_incomplete=1
+      continue
+    fi
+    if [[ ! -s "${scanner_json_file}" ]] \
```
dceoy, line 682: **P1 — Validate text and log evidence before reporting `pass`**

`classify_result` validates only each scanner's status file and JSON. If a scanner has status `0` plus valid JSON but its required `.txt` or `.log` is missing (or the text file is empty), this function returns `pass`; only the later `enforce` step detects the incomplete evidence. The uploaded `status.json` can therefore tell the controller that the target passed while the job correctly fails.

Apply the same required-evidence contract used by `enforce` before recording success: missing JSON/text/log/status, invalid JSON, or empty text should classify as `incomplete` (or `error` where appropriate). Add regression tests that remove a scanner `.txt` and `.log`, not only its JSON.

.github/actions/repository-security-scan/action.yml
```diff
@@ -16,6 +52,8 @@ runs:
...
```
dceoy: **P1 — Keep status finalization working when Aqua installation fails**

`if: (! cancelled())` makes the installer run after target setup failures, but it does not cover failure of the installer itself. If Aqua download/checksum/setup fails, `yq` may be unavailable; the later always-run `status` step then cannot generate `status.json`, so a scanner-dependency failure still loses the controller-facing error classification.

Make the minimal failure finalizer independent of Aqua-installed tools (or otherwise bootstrap its JSON writer separately) and incorporate installer failure into the recorded result. This path should still upload a valid identity-bound `error` status.

dceoy (commented):
Reviewed current head `5b71886cbcb9`.

Two blocking evidence/aggregation gaps remain. CI/CD is green, and the previously reported setup-finalization and identity issues are fixed, but `status.json` can still be absent or falsely report `pass` on required-evidence failures. I do not recommend merging until these are addressed.

.github/workflows/repository-security-scan.yml
```diff
@@ -1,13 +1,61 @@
...
+        description: GitHub App client ID used to mint a repository-scoped token for target-repository
+      TARGET_APP_PRIVATE_KEY:
+        required: false
+        description: GitHub App private key used to mint a repository-scoped token for target-repository
+concurrency:
+  # target-repository disambiguates concurrent matrix invocations from a
+  # periodic-scan controller; run_id is a safe fallback for any other
+  # workflow_call. Only pull_request runs cancel a superseded in-progress
+  # scan; merge_group and matrix target-mode runs must all complete.
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.event.merge_group.head_ref || inputs.target-repository || github.run_id }}
```
dceoy: **P1 — Prioritize the explicit target in the concurrency key**

When a trusted controller calls this reusable workflow from a `pull_request`-triggered matrix, `github.event.pull_request.number` wins before `inputs.target-repository`. Every target expansion therefore receives the same concurrency group, and `cancel-in-progress` is also `true`; the matrix jobs can cancel one another and omit the status/evidence that target mode promises. Key explicit target mode by `inputs.target-repository` first and disable PR cancellation when that input is set, e.g. make cancellation conditional on `inputs.target-repository == ''`.

dceoy (commented):
Found one correctness issue in target-mode concurrency. The current CI suite is green and all earlier review threads are resolved, but explicit target matrices can cancel one another when the controller is triggered by a pull request.

.github/workflows/repository-security-scan.yml
```diff
@@ -1,13 +1,61 @@
...
+        description: GitHub App client ID used to mint a repository-scoped token for target-repository
+      TARGET_APP_PRIVATE_KEY:
+        required: false
+        description: GitHub App private key used to mint a repository-scoped token for target-repository
+concurrency:
+  # target-repository disambiguates concurrent matrix invocations from a
+  # periodic-scan controller; run_id is a safe fallback for any other
+  # workflow_call. Only pull_request runs cancel a superseded in-progress
+  # scan; merge_group and matrix target-mode runs must all complete.
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ inputs.target-repository || github.event.pull_request.number || github.event.merge_group.head_ref || github.run_id }}
```
dceoy: **P1 — Keep every target-mode invocation in the concurrency queue**

Prioritizing `target-repository` prevents different repositories in one matrix from colliding, but repeated invocations for the same repository still share this group. With the default concurrency queue, GitHub retains only one running and one pending invocation; a third overlapping controller run cancels and replaces the existing pending scan even though `cancel-in-progress` is false. That drops the status/artifact required by the superseded controller run, contradicting the comment that matrix target-mode runs must all complete.

Include `github.run_id` (or another per-controller-run identity) in the target-mode group, or set `queue: max` if serialization across runs is intentional. Add a structural test that pins the chosen behavior.

.github/workflows/repository-security-scan.yml
```diff
@@ -33,14 +81,82 @@ jobs:
...
+          # target-repository/target-ref), those outputs are unset, and
+          # falling back to github.repository/github.sha would attribute
+          # the failed scan to the caller instead of the requested target.
+          # inputs.target-ref is placed last (not behind another `||`) so
+          # an empty-but-requested ref surfaces as empty evidence instead
+          # of silently falling through to github.sha: `A && '' || B`
+          # evaluates to B, since '' is falsy, so target-ref cannot sit on
+          # the truthy side of that pattern.
+          repository: ${{ inputs.target-repository == '' && github.repository || inputs.target-repository }}
+          commit-sha: ${{ inputs.target-repository == '' && (github.event_name == 'merge_group' && github.event.merge_group.head_sha || github.sha) || inputs.target-ref }}
```
dceoy: **P2 — Preserve the supplied ref in target-ref-only error evidence**

The resolver now correctly rejects `target-ref` without `target-repository`, but this expression selects `github.sha` whenever the repository input is empty. The always-run finalizer therefore emits an `error` status bound to the caller commit rather than the supplied invalid/requested ref (and may combine a target `repository-id` with the controller repository/SHA). That is the fallback the preceding comment says to avoid.

Select the raw `inputs.target-ref` whenever it is non-empty, independently of `target-repository`, and add a workflow-level regression test for the target-ref-only failure path.

dceoy (commented):
Reviewed current head `27cc9721341c`.

Two correctness gaps remain in explicit target mode. CI/CD is green and all earlier review threads are resolved, but overlapping controller runs can still lose matrix invocations, and the target-ref-only rejection records the wrong commit identity. I do not recommend merging until these are addressed.

.github/workflows/repository-security-scan.yml
cs-daichinarushima: Fixed in 20eb8b9: explicit target mode now keys concurrency by target repository plus caller run_id, with a structural regression test; target-mode cancellation remains disabled.

.github/workflows/repository-security-scan.yml
cs-daichinarushima: Fixed in 20eb8b9: finalizer evidence now uses raw inputs.target-ref when supplied without target-repository, with a structural regression test; the invalid request still cannot affect checkout.

dceoy (commented):
Reviewed current head `20eb8b91deda0d`.

One blocking correctness gap remains:

1. **P1 — Trusted evidence-conversion failures can still be recorded as policy findings.** `classify_result` treats every status `1` with valid JSON and no `.result == "not-run"` marker as `findings`. However, `run_actionlint` explicitly sets `scanner_status_value=1` when converting actionlint JSON Lines with `yq` fails, then replaces the JSON with `[]`. Once the text/log files exist, `status.json` therefore reports `findings` even though the trusted scanner evidence pipeline failed rather than detecting a target policy violation. Use an explicit per-scanner result/error marker, or assign a dedicated runtime-error status greater than `1`/`not-run` result for conversion and rendering failures. Add a regression test that forces the actionlint JSONL conversion path to fail.

Documentation note: README currently instructs users to configure this workflow directly as an organization ruleset required workflow immediately before explaining that the current `workflow_call`-only workflow cannot yet be selected that way. Make the first instruction conditional on Scope 1/direct-trigger activation.

CI/CD is green and the earlier setup-finalization, evidence-validation, identity, and concurrency issues are fixed. I do not recommend merging until the P1 classification gap is addressed.

dceoy:
Fixed in `54fff06`: actionlint JSONL conversion failures now use runtime status `2` and produce aggregate `error` evidence; regression coverage was added.

dceoy (commented):
Reviewed current head `54fff06e5fe2`.

One blocking classification gap remains:

1. **P1 — A reporting/conversion failure can be hidden when the primary scan already found issues.** `run_zizmor`, `run_actionlint`, `run_shellcheck`, and `run_trivy` replace the primary status with the renderer/converter status only when the primary status is `0`. If the JSON-producing scan returns `1` for findings and the subsequent plain-text render or `trivy convert` fails, the wrapper preserves status `1`; `ensure_text_evidence` then supplies a non-empty fallback, so `classify_result` records `findings` rather than `error`/`incomplete`. This violates the target-mode requirement to distinguish scanner runtime/evidence failures from policy findings and can make controller aggregation treat incomplete evidence as a valid findings result.

Preserve both execution outcomes explicitly. For example, promote any renderer/converter failure that is not the tool's normal findings exit to a dedicated runtime-error status/result, or record a per-scanner result marker consumed by `classify_result`. Add a regression test where the primary invocation returns findings and the secondary renderer/converter fails.

Non-blocking identity note: a `target-ref`-only invalid call currently records `repository: github.repository` together with the supplied foreign/invalid ref. Consider falling back to the event repository only when *both* target inputs are empty; otherwise leave repository empty so error evidence does not assert a repository/ref pair that was never requested or scanned.

CI/CD is green, and the previously reported setup-finalization, evidence-validation, concurrency, and actionlint JSONL-conversion issues are fixed. I do not recommend merging until the P1 mixed findings/runtime-failure path is covered.

dceoy (commented):
Reviewed current head `f3f155baf729`.

No blocking code findings remain. The latest change correctly promotes secondary renderer/converter failures over existing findings, the earlier setup-finalization/evidence/concurrency issues are resolved, and CI/CD is green.

Two non-blocking consistency issues remain:

1. **P2 — Do not pair a target-ref-only failure with the caller repository.** For an invalid call that supplies `target-ref` without `target-repository`, the finalizer preserves the supplied ref but still sets `repository` to `github.repository`. This can produce internally inconsistent identity evidence (for example, a target `repository-id`/default branch and foreign ref bound to the controller repository). Fall back to `github.repository` only when *both* target inputs are empty; otherwise leave repository empty when no target repository was supplied. Add a structural regression test for that exact expression.

2. **P3 — Make the ruleset instruction conditional on Scope 1.** README still first instructs users to configure this workflow directly as a `pull_request`/`merge_group` ruleset required workflow, then immediately states that the workflow exposes only `workflow_call` and cannot yet be selected that way. Rephrase the first paragraph as a future step after direct-trigger activation, or move it below the caveat.

These do not invalidate the implemented immutable-target happy path, but fixing them before merge would keep controller evidence and rollout documentation unambiguous.

dceoy (commented):
Reviewed current head `d3904b28fae`.

No blocking code findings remain. The immutable-target validation, trusted scanner pinning, setup-failure finalization, status classification, evidence identity, and target-mode concurrency fixes are internally consistent, all existing review threads are resolved, and CI/CD is green.

Two non-blocking PR-description corrections remain:

1. The Scope 2 interface bullet still names `TARGET_APP_ID`, while the workflow and README expose `TARGET_APP_CLIENT_ID` and pass it to the action's `client-id` input.
2. The description says `resolve-target.sh` is the single source for values used by both checkout and the composite-action call. The finalizer intentionally passes raw target inputs to the composite so resolution failures retain the requested identity; update that wording to distinguish checkout resolution from failure-evidence identity.

The disclosed limitation remains material: App-token scoping, private cross-repository checkout, and matrix artifact aggregation have not been exercised end to end. That should be verified during the downstream `segh` rollout, but it does not expose another static blocker in this PR.

PR #874, opened 2026-08-04
Inline supply-chain cooldown into local-qa script

dceoy:
## Summary
- Merge `.agents/skills/local-qa/scripts/supply-chain-cooldown.sh` into `qa.sh` and delete the separate helper.
- Update the local-qa skill install guidance to set cooldown env vars inline.
- Clear the repository-security-scan `source=` allowlist entry and remove the fixture that covered only that former pattern.

## Test plan
- [x] `./scripts/qa.sh`
- [ ] Confirm CI `repository-security-scan` / shell bats still pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/actions/repository-security-scan/scan.sh
+++ b/.github/actions/repository-security-scan/scan.sh
@@ -214,18 +214,11 @@ readonly shellcheck_directive_non_suppressing_keys=' shell '
 # directive key, remains rejected.
 readonly shellcheck_directive_trusted_disable_codes=' SC2016 SC2153 '
 
-# source= values that are centrally vetted as accurately identifying the
-# real file a dynamic `source` statement resolves to at that exact line
-# (unlike shell=, source= can misdirect analysis, so - like disable= - only
-# specific, reviewed values are trusted, not the key unconditionally).
-# Currently empty: no target-owned source= directives are trusted.
-readonly shellcheck_directive_trusted_source_paths=' '
-
 # Classifies one already-matched "# shellcheck key=value ..." line (with any
 # leading line-number/filename prefix already stripped by the caller).
 # Returns success (0) when the line is a policy violation, failure (1) when
-# every key=value token on it is a non-suppressing key, an explicitly
-# trusted disable= code, or an explicitly trusted source= path.
+# every key=value token on it is a non-suppressing key or an explicitly
+# trusted disable= code.
 directive_line_is_policy_violation() {
   local directive_line="$1"
   local directive_tail token key value code
@@ -249,12 +242,6 @@ directive_line_is_policy_violation() {
             *) return 0 ;;
           esac
         done
-        ;;
-      source)
-        case "${shellcheck_directive_trusted_source_paths}" in
-          *" ${value} "*) ;;
-          *) return 0 ;;
-        esac
         ;;
       *) return 0 ;;
     esac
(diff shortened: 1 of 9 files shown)


.github/actions/repository-security-scan/scan.sh
```diff
@@ -218,10 +218,8 @@ readonly shellcheck_directive_trusted_disable_codes=' SC2016 SC2153 '
 # real file a dynamic `source` statement resolves to at that exact line
 # (unlike shell=, source= can misdirect analysis, so - like disable= - only
 # specific, reviewed values are trusted, not the key unconditionally).
-# Currently only .agents/skills/local-qa/scripts/qa.sh's source of its
-# sibling supply-chain-cooldown.sh, which shellcheck cannot follow on its
-# own because the path is built from a runtime-computed REPO_ROOT.
-readonly shellcheck_directive_trusted_source_paths=' .agents/skills/local-qa/scripts/supply-chain-cooldown.sh '
+# Currently empty: no target-owned source= directives are trusted.
+readonly shellcheck_directive_trusted_source_paths=' '
```
dceoy: **P2 — Remove the now-unused `source=` exception path.** With this allowlist intentionally empty and its positive fixture deleted, the classifier can no longer legitimately accept any `source=` directive, so this generic exception mechanism is now unreachable and its acceptance behavior is untested. Because this code defines a security boundary, prefer treating `source)` as an unconditional policy violation (and removing this constant plus the related classifier wording). Reintroduce a narrowly tested allowlist only when a concrete trusted use case exists.

dceoy (commented):
The inline cooldown refactor preserves the existing values and removes the only repository-owned need for a `source=` ShellCheck directive. CI/CD and Repository security are green. I found one non-blocking security/maintainability simplification: remove the now-unused `source=` exception mechanism rather than retaining an unreachable, untested allowlist path.

.github/workflows/opencode-bot.yml
```diff
@@ -105,7 +105,7 @@ jobs:
           token: ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}  # zizmor: ignore[secrets-outside-env] caller-provided secret
           persist-credentials: false
       - name: Run OpenCode
-        uses: dceoy/opencode-action@7778b833178318ade66f12d42102dd9eb226bfb2  # v0.6.3
+        uses: dceoy/opencode-action@da47df8f9d60c12de7b76dc1ca37633b147f0241  # v0.6.4
```
dceoy, line 108: **P2 — Split or document the unrelated OpenCode Action upgrade.** This PR is scoped to the local-QA cooldown and ShellCheck directive policy, but it also upgrades `dceoy/opencode-action` from v0.6.3 to v0.6.4 in both bot workflows. The PR summary does not mention the dependency change, and the stated test plan only covers local QA and the repository-security scan, so the workflow behavior introduced by the new action revision is not reviewed or exercised here. Prefer moving both pin updates to a dedicated dependency PR; alternatively, add them to the summary and validate the bot/review paths explicitly.

dceoy (commented):
The cooldown inlining and unconditional `source=` rejection are internally consistent, and the latest CI/CD and Repository security runs pass. The previous `source=` allowlist finding is resolved. I found one non-blocking scope/testability issue: the PR also upgrades `opencode-action` in two workflows without documenting or validating that behavior change.

PR #908, opened 2026-08-11
Add reusable Markdown lint workflow

dceoy:
## Summary
- add a read-only reusable `markdown-lint.yml` workflow using pinned `markdownlint-cli2`
- add `use-markdownlint-cli2`, `use-prettier`, and `use-biome` inputs so callers can select Markdown checks independently
- run `prettier --check` against the tracked Markdown file set and optionally run Biome formatting checks
- expose `markdownlint-config-data` so callers can control the inline `markdownlint-cli2` base configuration without a repository config file
- keep the default markdownlint behavior equivalent to `MD013` disabled
- skip tracked Markdown symlinks so explicit formatter arguments remain valid
- pin Prettier and Biome in `markdown-format-and-pr.yml` with Renovate annotations instead of installing floating npm versions
- add the Markdown lint workflow to the repository CI lint gate and Dependabot auto-merge prerequisites

## Notes
- YAML changes intentionally contain no blank lines
- `markdownlint-config-data` is validated with `jq`, written to a temporary `.jsonc` config, and passed through `markdownlint-cli2 --config`
- Biome currently does not support Markdown parsing/formatting (confirmed against the pinned 2.5.5 configuration schema: no `markdown` key exists), so the Biome path uses only `--files-ignore-unknown=true`; it fails closed with a nonzero exit when no Markdown files are processed instead of silently succeeding, so enabling `use-biome` for Markdown is a no-op that surfaces as a failure until Biome adds Markdown support upstream
- Biome is pinned to 2.5.5 to respect the repository Renovate `minimumReleaseAge` policy; Prettier is pinned to 3.9.6 and markdownlint-cli2 to 0.23.2
- README workflow documentation remains generated by the existing `update-readme-md` CI job

## Validation
- exercised `use-biome: true` in GitHub Actions with no Markdown files matched by Biome: install, markdownlint-cli2, and `prettier --check` steps passed, and the Biome step now fails closed (nonzero exit) rather than silently succeeding, since Biome 2.5.5 has no Markdown support
- restored repository CI to the workflow defaults after the Biome-path validation
- GitHub Actions lint/scan, yamllint, ShellCheck, README generation, and Markdown formatting jobs remain part of the PR checks


.github/workflows/markdown-lint.yml
```diff
@@ -0,0 +1,159 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020  # v7.0.0
+        with:
+          node-version: ${{ inputs.node-version }}
+      - name: Cache Markdown lint tools
+        if: inputs.enable-cache
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9  # v6.1.0
+        with:
+          path: ~/.cache/gha-for-devops/markdown-lint
+          key: markdown-lint-${{ runner.os }}-${{ inputs.node-version }}-${{ inputs.use-markdownlint-cli2 }}-${{ inputs.use-prettier }}-${{ inputs.use-biome }}-${{ env.MARKDOWNLINT_CLI2_VERSION }}-${{ env.PRETTIER_VERSION }}-${{ env.BIOME_VERSION }}-${{ inputs.cache-salt }}
```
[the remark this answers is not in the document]
dceoy: Cache key now includes `runner.arch` (2cf5861) so restoring across architectures can't reuse Biome's arch-specific optional binary.

.github/workflows/markdown-lint.yml
```diff
@@ -0,0 +1,159 @@
...
+          USE_PRETTIER: ${{ inputs.use-prettier }}
+          USE_BIOME: ${{ inputs.use-biome }}
+        run: | # zizmor: ignore[adhoc-packages] installs pinned Markdown lint tools
+          packages=()
+          needs_install=false
+          tool_dir="${HOME}/.cache/gha-for-devops/markdown-lint"
+          if ${USE_MARKDOWNLINT_CLI2}; then
+            [[ "${MARKDOWNLINT_CLI2_VERSION}" =~ ^[0-9]+(\.[0-9]+)*([._+-][A-Za-z0-9]+)*$ ]] || { echo "invalid MARKDOWNLINT_CLI2_VERSION"; exit 1; }
+            packages+=("markdownlint-cli2@${MARKDOWNLINT_CLI2_VERSION}")
+            [[ -x "${tool_dir}/node_modules/.bin/markdownlint-cli2" ]] || needs_install=true
```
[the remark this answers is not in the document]
dceoy: Install check now compares the installed package.json version to the pinned version instead of just checking for an executable, so a stale install left on a persistent runner no longer skips a version bump (2cf5861).

.github/workflows/markdown-format-and-pr.yml
```diff
@@ -85,7 +100,7 @@ jobs:
         env:
           BIOME_GLOB: ${{ inputs.biome-glob }}
         run: >
-          biome format --write --files-ignore-unknown=true "${BIOME_GLOB}"
+          biome format --write --files-ignore-unknown=true --no-errors-on-unmatched "${BIOME_GLOB}"
```
dceoy: **[P2] Don't turn Biome-only mode into a successful no-op**

Biome 2.5.5 includes an experimental Markdown formatter, but it is disabled by default (`MarkdownFormatterEnabled = Bool<false>`). This workflow never enables it. Adding `--no-errors-on-unmatched` therefore lets `use-prettier: false, use-biome: true` succeed without formatting any Markdown; the new `markdown-lint.yml` has the same false-green behavior. Either explicitly enable the Markdown formatter (for example via a temporary Biome config) or reject/remove Biome-only mode until it is actually enforced.

dceoy (commented):
Found one correctness issue; see the inline comment.

.github/workflows/markdown-format-and-pr.yml
dceoy: Agreed this was a false-green risk, but confirmed the fix has to be different from enabling a config knob: the pinned Biome 2.5.5 (`BIOME_VERSION` in these workflows) has no `markdown` key in its configuration schema at all — `markdown.formatter.enabled` doesn't exist in this version, so there's no way for a caller to opt in via config. I verified this directly: `biome format --files-ignore-unknown=true test.md` against the pinned 2.5.5 exits 1 with "No files were processed in the specified paths" once `--no-errors-on-unmatched` is removed, and `grep -i markdown` over the installed `configuration_schema.json` returns nothing.

Fix pushed in a91f2ad: dropped `--no-errors-on-unmatched` from the Biome step in both `markdown-format-and-pr.yml` and the new `markdown-lint.yml`, restoring the fail-closed behavior `markdown-format-and-pr.yml` had before this PR (it never had that flag on `main`). `use-biome: true` now fails instead of silently succeeding when Biome can't process the Markdown glob, until Biome adds real Markdown support upstream. PR description's Notes/Validation sections updated to match.

.github/workflows/markdown-lint.yml
```diff
@@ -0,0 +1,162 @@
...
+        description: Path to search for Markdown files
+        default: .
+      file-names:
+        required: false
+        type: string
+        description: Newline-separated file names or pathspecs passed to git ls-files
+        default: |
+          *.md
+          *.markdown
+          *.mdx
```
dceoy, line 18: **[P2] Avoid linting MDX with markdownlint-cli2 by default**

`file-names` includes `*.mdx` while `use-markdownlint-cli2` defaults to `true`. markdownlint-cli2 targets Markdown/CommonMark and its documented default file patterns are Markdown extensions, not MDX; valid MDX JSX/ESM can therefore be interpreted as ordinary Markdown/HTML and trigger lint rules. A caller using this workflow unchanged can fail on otherwise valid MDX. Please keep MDX out of the markdownlint input by default (for example, split the markdownlint and formatter file sets so Prettier can still check MDX).

dceoy (commented):
Reviewed the current head. One inline finding remains.

.github/workflows/markdown-lint.yml
dceoy: Confirmed empirically against the pinned toolchain: markdownlint-cli2 v0.23.2 parses `<Chart data={data} />` as MD033 inline HTML and flags a leading `import` line as a missing top-level heading (MD041), while pinned Prettier 3.9.6 correctly reformats the same MDX file's JS/JSX (adds semicolons/quotes, leaves the JSX and expression untouched). So MDX support is asymmetric between the two tools, exactly as described.

Fixed in 976a76a: added a `markdownlint-file-names` input (defaulting to `*.md`/`*.markdown`, no `.mdx`) used only by the new "Create markdownlint-cli2 file list" step feeding `markdownlint-cli2`. The existing `file-names` input (still defaulting to `*.md`/`*.markdown`/`*.mdx`) continues to feed Prettier and Biome unchanged. Callers who do want markdownlint-cli2 to lint MDX can opt in explicitly via `markdownlint-file-names`.

dceoy (commented):
Reviewed the current head (`976a76a`). No additional code findings. The prior inline findings are addressed and resolved, and CI/CD passes. The remaining `Repository security` failure appears unrelated to this PR: actionlint emitted no findings, while its embedded ShellCheck invocation through Aqua failed during checksum metadata parsing with `EOF`.

PR #909, opened 2026-08-12
Use npx and uvx for local QA tools

dceoy:
## Summary
- run unlocked Ruff, Pyright, and Python-distributed QA tools through `uvx`
- run applicable JavaScript/TypeScript QA tools through `npx`
- use the explicit `typescript` package when invoking `tsc`
- run `markdownlint-cli2 --fix` for tracked Markdown files, using `.markdownlint-cli2.jsonc` when present and a temporary config disabling `MD013`, `MD033`, and `MD041` otherwise
- keep project-locked `uv run`/Poetry paths and native tools such as `actionlint`, `shellcheck`, `shfmt`, and `trivy` unchanged
- retain the existing 7-day uv/npm/pnpm package cooldown

## Rationale
This aligns the generic local QA runner with the `opencode-action` package-runner pattern and the Markdown lint behavior introduced from `oracle-pr-loop`, while preserving project lockfile behavior. Package-on-demand Python tools honor `UV_EXCLUDE_NEWER`, and package-on-demand Node.js tools honor `NPM_CONFIG_MIN_RELEASE_AGE`, instead of depending on globally installed versions.

### DIFF
--- a/.agents/skills/local-qa/scripts/qa.sh
+++ b/.agents/skills/local-qa/scripts/qa.sh
@@ -103,7 +103,7 @@ if [[ "${N_MARKDOWN_FILES}" -gt 0 ]]; then
   if [[ -f .markdownlint-cli2.jsonc ]]; then
     git ls-files -z -- '*.md' '*.mdx' | xargs -0 -t npx -y markdownlint-cli2 --fix --config .markdownlint-cli2.jsonc
   else
-    printf '{"config":{"MD013":false}}' > .markdownlint-cli2.jsonc
+    printf '{"config":{"MD013":false,"MD033":false,"MD041":false}}' > .markdownlint-cli2.jsonc
     set +e
     git ls-files -z -- '*.md' '*.mdx' | xargs -0 -t npx -y markdownlint-cli2 --fix --config .markdownlint-cli2.jsonc
     markdownlint_exit_code="${?}"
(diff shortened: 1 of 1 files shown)


.agents/skills/local-qa/scripts/qa.sh
```diff
@@ -96,7 +96,20 @@ else
...
     npx -y prettier --write './**/*.{md,mdx}'
-    # markdownlint-cli2 './**/*.md'
+  fi
+fi
+
+if [[ "${N_MARKDOWN_FILES}" -gt 0 ]]; then
+  if [[ -f .markdownlint-cli2.jsonc ]]; then
+    git ls-files -z -- '*.md' '*.mdx' | xargs -0 -t npx -y markdownlint-cli2 --fix --config .markdownlint-cli2.jsonc
+  else
+    printf '{"config":{"MD013":false}}' > .markdownlint-cli2.jsonc
```
dceoy: **P2: Preserve other supported markdownlint-cli2 config formats**

This fallback only recognizes `.markdownlint-cli2.jsonc`. If a target repository uses another supported CLI2 config such as `.markdownlint-cli2.yaml`, `.markdownlint-cli2.cjs`, or `.markdownlint-cli2.mjs`, this branch creates a temporary `.markdownlint-cli2.jsonc`; markdownlint-cli2 gives that filename higher precedence, so the repository's existing CLI2 configuration is masked. That can drop project-specific `globs`, `overrides`, custom rules/plugins, etc. Detect the other supported CLI2 config names (or otherwise preserve normal config discovery) before installing the MD013 fallback.

dceoy (commented):
Found one additional issue in the Markdown-lint path added after the prior review.

PR #913, opened 2026-08-14
Remove repository security implementation

dceoy:
## Summary

- Remove the repository security scan reusable workflow and composite action.
- Remove repository-security-specific configuration, tests, fixtures, and documentation.
- Keep `gha-for-devops` focused on its reusable DevOps workflow scope.

## Rationale

Repository security scanning is outside the intended scope of this repository, so the dedicated implementation and its supporting surface should be removed rather than maintained here.

## Verification

- The change removes the repository-security implementation and its associated support files/documentation.
- Unrelated reusable workflows remain unchanged.


.github/workflows/repository-security-scan.yml
```diff
@@ -1,156 +1,47 @@
...
-          # empty ref retains an empty commit-sha.
-          repository: ${{ inputs.target-repository == '' && inputs.target-ref == '' && github.repository || inputs.target-repository }}
-          commit-sha: ${{ inputs.target-repository == '' && inputs.target-ref == '' && (github.event_name == 'merge_group' && github.event.merge_group.head_sha || github.sha) || inputs.target-ref }}
-          evidence-artifact-name: ${{ inputs.evidence-artifact-name }}
-          setup-failed: ${{ steps.target.outcome == 'failure' || steps.target-app-token.outcome == 'failure' || steps.target-checkout.outcome == 'failure' }}
+          scan-type: fs
+          scan-ref: .
+          scanners: vuln,secret,misconfig
+          severity: UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
+          ignore-unfixed: true
```
dceoy: [P1] Preserve the existing vulnerability gate semantics here. The removed scanner ran the vulnerability check at `HIGH,CRITICAL` and did not ignore unfixed findings; this combined Trivy step uses all severities and `ignore-unfixed: true`. That simultaneously starts failing on LOW/MEDIUM fixed vulnerabilities while allowing unfixed HIGH/CRITICAL vulnerabilities that previously blocked the gate. The PR rationale scopes out orchestration/evidence behavior, but does not describe this vulnerability-policy change. Please preserve the prior vuln semantics (for example, split vuln from secret/misconfig scans) or make the policy change explicit and justified.

dceoy (commented):
I found one blocking semantic regression in the Trivy replacement; see the inline comment.

.github/workflows/repository-security-scan.yml
```diff
@@ -1,156 +1,55 @@
...
   contents: read
 jobs:
-  scan:
-    name: scan
+  github-actions:
+    permissions:
+      contents: read
+      security-events: write
+      actions: read
+    uses: ./.github/workflows/github-actions-lint-and-scan.yml
```
dceoy: [P1] Keep Checkov policy outside the scanned revision. `github-actions-lint-and-scan.yml` leaves `checkov-config-file` unset, so Checkov auto-discovers `.checkov.yaml` / `.checkov.yml` from the scanned repository. The removed implementation always supplied the trusted config explicitly. With this replacement, a pull request can add or change `.checkov.yml` (for example `skip-check`) and suppress findings in the same security gate that is supposed to evaluate that PR. Please force a trusted/empty config here (or in the generic reusable workflow) rather than allowing target-owned Checkov policy to affect this gate.

.github/workflows/repository-security-scan.yml
```diff
@@ -1,156 +1,55 @@
...
+    with:
+      search-path: .github/workflows
+      use-pyflakes: false
+      use-yamllint: false
+      zizmor-search-path: .github
+      zizmor-advanced-security: false
+  shell:
+    permissions:
+      contents: read
+    uses: ./.github/workflows/shell-lint.yml
```
dceoy: [P1] Preserve ShellCheck coverage for extensionless shell scripts. The removed scanner inspected every tracked regular file's shebang and scanned shell interpreters even when the filename had no `.sh`/`.bash`/`.bats` suffix; the deleted fixtures explicitly covered extensionless and `/usr/bin/env` shebang cases. `shell-lint.yml` defaults to only `*.sh`, `*.bash`, and `*.bats`, so common executable scripts such as `bin/tool` now bypass this security gate. Please move shebang-based discovery into the generic shell workflow (or otherwise preserve that coverage) rather than dropping it as part of this refactor.

.github/workflows/repository-security-scan.yml
```diff
@@ -1,156 +1,55 @@
...
-      - name: Mint scoped token for explicit target repository
-        # Only a trusted organization controller supplies target-repository
-        # via workflow_call; direct pull_request/merge_group triggers never
-        # set it, so this step is skipped and github.token scans the caller
-        # repository unchanged.
-        if: inputs.target-repository != ''
-        id: target-app-token
-        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1  # v3.2.0
+      - name: Run Trivy vulnerability scan
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25  # v0.36.0
```
dceoy: [P1] Do not let the scanned revision supply Trivy ignore policy. The removed scanner invoked Trivy with `--config /dev/null --ignorefile /dev/null`; this action invocation does not override Trivy's default `.trivyignore`, so a PR can add `.trivyignore` and suppress vulnerability findings. The same applies to the secret/misconfiguration invocation below. Please force a trusted/empty ignore file for both scans so the gate remains fail-closed with respect to target-owned suppression files.

dceoy (commented):
The simplification is directionally good, and the previous Trivy severity regression is fixed. I found remaining security-gate regressions where the replacement now trusts suppression policy from the repository being scanned or reduces the set of shell scripts covered. These should be addressed before merge without reintroducing the deleted orchestration/evidence machinery.

.github/workflows/repository-security-scan.yml
dceoy: Fixed in 0683480: the repository security wrapper now passes `checkov-config-file: /dev/null`, so Checkov cannot auto-load target-owned `.checkov.yml` policy.

.github/workflows/repository-security-scan.yml
dceoy: Fixed in 429e043: `shell-lint.yml` now preserves the configured pathspecs and additionally discovers tracked regular files whose shebang selects a supported shell, including extensionless and `/usr/bin/env` scripts.

.github/workflows/repository-security-scan.yml
dceoy: Fixed in 0683480: both Trivy invocations now force `TRIVY_IGNOREFILE=/dev/null`, preventing target-owned `.trivyignore` files from suppressing findings.

.github/workflows/repository-security-scan.yml
```diff
@@ -1,156 +1,58 @@
...
-        uses: ./_trusted/.github/actions/repository-security-scan
+          scan-type: fs
+          scan-ref: .
+          scanners: vuln
+          severity: HIGH,CRITICAL
+          exit-code: '1'
+        env:
+          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db
+          TRIVY_JAVA_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-java-db
+          TRIVY_IGNOREFILE: /dev/null
```
dceoy: [P1] Disable repository-owned Trivy config, not just `.trivyignore`. Trivy defaults `--config` to `trivy.yaml`, and `trivy-action` only overrides that when `trivy-config` / `TRIVY_CONFIG` is set. A scanned PR can therefore add `trivy.yaml` and change options that are not explicitly overridden here (for example `scan.skip-dirs`, `scan.skip-files`, `ignore-policy`, or scanner-specific settings) to suppress findings. The removed scanner explicitly used `--config /dev/null`. Please pass `trivy-config: /dev/null` (or set `TRIVY_CONFIG: /dev/null`) for both Trivy invocations so target-owned policy cannot weaken this gate.

dceoy (commented):
Found one remaining P1 security-boundary issue. CI is green, and the earlier Checkov / `.trivyignore` / shell-discovery findings are addressed, but Trivy's repository-owned config is still trusted.

.github/workflows/repository-security-scan.yml
dceoy: Fixed in 6c8091a: both Trivy invocations now pass `trivy-config: /dev/null`, so repository-owned `trivy.yaml` cannot alter the security gate policy.

.github/workflows/repository-security-scan.yml
```diff
@@ -1,156 +1,60 @@
...
   contents: read
 jobs:
-  scan:
-    name: scan
+  github-actions:
+    permissions:
+      contents: read
+      security-events: write
+      actions: read
+    uses: ./.github/workflows/github-actions-lint-and-scan.yml
```
dceoy: [P1] Keep Actionlint policy outside the scanned revision as well. This nested workflow invokes `actionlint` without an explicit config path, and Actionlint automatically consumes `.github/actionlint.yaml` / `.github/actionlint.yml`; its `paths.*.ignore` entries can filter arbitrary findings. A PR being scanned can therefore add or modify that file and suppress Actionlint (including embedded ShellCheck) errors, even though this wrapper already forces trusted/empty Checkov and Trivy policy. Please add an explicit Actionlint config input to the generic workflow and pass a trusted/empty config here.

.github/workflows/shell-lint.yml
```diff
@@ -78,9 +78,25 @@ jobs:
...
+          while IFS= read -r -d '' entry; do
+            mode=${entry%% *}
+            [[ "${mode}" == '100644' || "${mode}" == '100755' ]] || continue
+            file=${entry#*$'\t'}
+            first_line=''
+            IFS= read -r first_line < "${file}" || true
+            [[ "${first_line}" =~ ${shell_shebang} ]] && selected["${file}"]=1
+          done < <(git ls-files -z --stage)
+
+          ((${#selected[@]} == 0)) || printf '%s\0' "${!selected[@]}" | xargs -0 -r -t shellcheck --external-sources
```
dceoy: [P1] Terminate ShellCheck options before appending target-controlled filenames. The new shebang discovery can select an extensionless tracked file named, for example, `--exclude=SC2086`; because `xargs` currently appends filenames directly after `shellcheck --external-sources`, ShellCheck interprets that filename as a CLI option and the scanned revision can suppress findings in other scripts. The removed implementation passed `--` before its file list. Please invoke ShellCheck as `shellcheck --external-sources -- <files...>` (equivalently place `--` in the xargs command before appended filenames).

dceoy (commented):
I found two remaining bypasses in the simplified security gate. Both let the scanned revision influence or weaken scanner behavior and should be fixed before merge.

dceoy (commented):
Blocking findings remain around target-controlled scanner suppression. GitHub does not allow requesting changes on one's own pull request, so these are posted as review comments instead.

.github/workflows/repository-security-scan.yml
```diff
@@ -1,156 +1,61 @@
...
+    with:
+      search-path: .github/workflows
+      actionlint-config-file: /dev/null
+      use-pyflakes: false
+      use-yamllint: false
+      zizmor-search-path: .github
```
dceoy: [P1] Preserve zizmor's target-policy isolation. The removed scanner invoked zizmor with both `--no-config` and `--no-ignores`; the generic workflow now calls `zizmor-action` without either guarantee. The pinned action only adds `--config` when a config input is supplied and exposes no `--no-ignores` switch, so a scanned revision can again use repository-owned zizmor configuration / ignore annotations to suppress findings. Please keep config and ignores outside the scanned revision (or stop advertising this wrapper as a trusted security gate).

.github/workflows/repository-security-scan.yml
```diff
@@ -1,156 +1,61 @@
...
+      zizmor-advanced-security: false
+      checkov-config-file: /dev/null
```
dceoy: [P1] `/dev/null` only prevents Checkov from auto-loading repository config; it does not disable inline `checkov:skip=` suppressions. The removed scanner explicitly inspected `skipped_checks` and failed when a target revision used inline suppression, and this PR deletes the fixture that covered that boundary. As written, a PR can suppress the exact Checkov finding this gate is intended to evaluate. Please preserve the prior fail-closed handling for inline skips, or make the remaining workflow non-security/non-enforcing.

.github/workflows/shell-lint.yml
```diff
@@ -78,9 +78,25 @@ jobs:
...
+          while IFS= read -r -d '' entry; do
+            mode=${entry%% *}
+            [[ "${mode}" == '100644' || "${mode}" == '100755' ]] || continue
+            file=${entry#*$'\t'}
+            first_line=''
+            IFS= read -r first_line < "${file}" || true
+            [[ "${first_line}" =~ ${shell_shebang} ]] && selected["${file}"]=1
+          done < <(git ls-files -z --stage)
+
+          ((${#selected[@]} == 0)) || printf '%s\0' "${!selected[@]}" | xargs -0 -r -t shellcheck --external-sources --
```
dceoy: [P1] Keep ShellCheck policy outside the scanned revision. This invocation does not pass a trusted `--rcfile` (and the Actionlint path likewise no longer sets trusted `SHELLCHECK_OPTS`), so ShellCheck can consume repository-owned `.shellcheckrc` settings and suppress diagnostics. The removed scanner explicitly passed the trusted rcfile and rejected target-owned suppression directives. Please force a trusted/empty rcfile for the security path and retain rejection of target-owned suppressing directives, or otherwise avoid treating this as a fail-closed security gate.

dceoy (commented):
No blocking findings. The final diff cleanly removes the repository-security-specific workflow, composite action, configuration, tests/fixtures, and documentation without modifying unrelated reusable workflows. CI/CD is green on the current HEAD.

PR #916, opened 2026-08-14
Migrate repository tooling to mise

dceoy:
## Summary

- remove the repository devcontainer configuration
- replace Aqua with `mise.toml` for repository tool versions
- use SHA-pinned `jdx/mise-action` in repository CI
- remove the Aqua-specific bootstrap script and Renovate preset
- let Renovate's native `mise` manager update tools declared in `mise.toml`

## Dependency update ownership

Dependabot does not currently expose a `mise` package ecosystem, so it continues to manage `jdx/mise-action` through the existing `github-actions` ecosystem without mise-specific grouping. Tool versions inside `mise.toml` are managed by Renovate's native `mise` manager.

## Validation

- confirmed the branch is based directly on current `main` with only the intended 8-file diff
- validated `mise.toml` as TOML, `.github/renovate.json` as JSON, and `.github/dependabot.yml` as YAML
- preserved the exact tool versions previously pinned in `aqua.yaml`
- pinned `jdx/mise-action` v4.2.3 to commit `9e7f7633ff6f6d6048a9418a68d48f288f50eb14`


.github/workflows/ci.yml
```diff
@@ -66,13 +66,8 @@ jobs:
...
           persist-credentials: false
-      - name: Install aqua and CLI tools
-        uses: aquaproj/aqua-installer@96a9bc20066c5bf5e275b41019cfc165b25f4e2e  # v4.0.5
-        with:
-          aqua_version: v2.60.2
-          aqua_opts: ""
-      - name: Verify yq and gomplate are installed
-        run: ./.github/scripts/install-yq-gomplate.sh
+      - name: Set up mise
+        uses: jdx/mise-action@9e7f7633ff6f6d6048a9418a68d48f288f50eb14  # v4.2.3
```
dceoy, line 73: **P1 — Pin the mise executable version.** `jdx/mise-action` documents that omitting `with.version` installs the latest mise release. The old Aqua setup pinned `aqua_version: v2.60.2`, so after this change the installer/runtime (including its bundled registry/backend behavior) can change between CI runs without any repository diff. Add an explicit mise version here and to the second invocation; ideally annotate/manage that pin with Renovate so upgrades remain deliberate.

dceoy (commented):
Found one blocking reproducibility issue: the migration pins the action SHA and tool versions, but leaves the mise executable itself floating at `latest`. Please pin the mise version in both `jdx/mise-action` steps (and preferably make that version Renovate-managed).

PR #928, opened 2026-08-17
Propagate release lookup failures

dceoy:
## Summary
- resolve the latest release tag in a standalone assignment
- only write `LATEST_TAG` to `GITHUB_ENV` after the GitHub CLI command succeeds

## Root cause
`gh release list` ran inside a command substitution embedded in a `tee` here-string. When GitHub returned HTTP 503, `tee` still exited successfully, so the workflow continued with an empty tag and incorrectly reported `No latest release found`.

## Validation
- verified under `bash -euo pipefail` that a standalone assignment preserves a failing command substitution exit status
- branch diff is limited to the release lookup step in `.github/workflows/github-major-version-tag.yml`

### DIFF
--- a/.github/workflows/github-major-version-tag.yml
+++ b/.github/workflows/github-major-version-tag.yml
@@ -49,10 +49,11 @@ jobs:
         env:
           GITHUB_REPOSITORY: ${{ github.repository }}
         run: |
-          tee -a "${GITHUB_ENV}" <<< "LATEST_TAG=$(
+          LATEST_TAG="$(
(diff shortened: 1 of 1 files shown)


dceoy (commented):
Reviewed the release lookup failure-propagation fix. No actionable issues found. Moving `gh release list` into a standalone assignment makes the assignment inherit the command substitution's non-zero status, so `bash -euo pipefail` stops the step before writing an empty `LATEST_TAG` to `GITHUB_ENV`. CI/CD passes on the reviewed head commit.

